PR #365, opened 2021-11-24
Update README.md

asifalix:


asifalix:
Removing all references to calico, bastion etc..

PR #386, opened 2022-01-27
Add createTime and updateTime in db for systempolicy

seswarrajan:

### DIFF
--- a/src/libs/mysqlHandler.go
+++ b/src/libs/mysqlHandler.go
@@ -497,8 +497,8 @@ func CreateTableWorkLoadProcessFileSetMySQL(cfg types.ConfigDB) error {
 			"	`fromSource` varchar(256) DEFAULT NULL," +
 			"	`settype` varchar(16) DEFAULT NULL," + // settype: "file" or "process"
 			"	`fileset` text DEFAULT NULL," +
-			"	`createdTime` int NOT NULL," +
-			"	`updatedTime` int NOT NULL," +
+			"	`createdTime` bigint NOT NULL," +
+			"	`updatedTime` bigint NOT NULL," +
 			"	PRIMARY KEY (`id`)" +
 			"  );"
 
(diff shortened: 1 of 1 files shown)


src/libs/mysqlHandler.go
```diff
@@ -566,6 +568,7 @@ func GetWorkloadProcessFileSetMySQL(cfg types.ConfigDB, wpfs types.WorkloadProce
 	var fs []string
 	var policyNames []string
 	var policyName string
+	var createdTime, updatedTime int64
```
nyrahul: why is this `int64` when the vars in schema are `int`.

nyrahul (changes requested):
mostly LGTM. Just one inline comment. Pls chk.

src/libs/mysqlHandler.go
```diff
@@ -521,7 +523,7 @@ func GetWorkloadProcessFileSetMySQL(cfg types.ConfigDB, wpfs types.WorkloadProce
 	var results *sql.Rows
 	var err error
 
-	query := "SELECT policyName,clusterName,namespace,containerName,labels,fromSource,settype,fileset FROM " + WorkloadProcessFileSet_TableName
+	query := "SELECT policyName,clusterName,namespace,containerName,labels,fromSource,settype,fileset,createdtime,updatedtime FROM " + WorkloadProcessFileSet_TableName
```
nyrahul:
```suggestion
	query := "SELECT policyName,clusterName,namespace,containerName,labels,fromSource,settype,fileset FROM " + WorkloadProcessFileSet_TableName
```

nyrahul: we do not need to get `createdTime`, `updatedTime` ... we are not using it.

src/libs/mysqlHandler.go
nyrahul:
```suggestion
```

src/libs/mysqlHandler.go
```diff
@@ -577,6 +580,8 @@ func GetWorkloadProcessFileSetMySQL(cfg types.ConfigDB, wpfs types.WorkloadProce
 			&loc_wpfs.FromSource,
 			&loc_wpfs.SetType,
 			&fscsv,
+			&createdTime,
+			&updatedTime,
```
nyrahul:
```suggestion
```

nyrahul (changes requested):

nyrahul (approved):

PR #388, opened 2022-01-29
Added new gRPC method to fetch system observability data

seswarrajan:


scripts/get_sys_observability_data.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
nyrahul, line 1: we don't want the input to be provided in this form (i.e., do not use stdin to get input)...

Read the parameters from commandline...

`./get_sys_observability_data.sh --cluster default --container xyz --namespace abc --labels "xyz=abc, lmn=pqr"`

scripts/get_sys_observability_data.sh
```diff
@@ -0,0 +1,20 @@
...
+echo "Input containername"
+read containername
+
+echo "Input namespace"
+read namespace
+
+echo "Input labels"
+read labels
+
+DATA='{"clusterName": "'$clustername'", "namespace":"'$namespace'", "containerName":"'$containername'", "labels":"'$labels'"}'
```
nyrahul, line 20: Is it necessary that all the parameters are provided?

If no clustername is provided .. then we should be using all clusters ... Similarly for other fields.

src/libs/common.go
```diff
@@ -452,6 +462,34 @@ func WriteKubeArmorPolicyToYamlFile(fname string, namespace string, policies []t
 	}
 }
 
+func WriteSysObsDataToJsonFile(obsData types.SysObservabilityData) {
```
nyrahul, line 465: Send the output as response to request ... Not dump it in the file.

src/systempolicy/observability.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil
+	}
+
+	var sysObsProcessFileData []types.SysObsProcessFileData
+
+	for wpfs, fsset := range wpfsSet {
+		var locSysObsProcessFileData types.SysObsProcessFileData
+
+		locSysObsProcessFileData.FromSource = wpfs.FromSource
+		if wpfs.SetType == "Process" {
```
nyrahul:
```suggestion
		if wpfs.SetType == SYS_OP_PROCESS {
```

src/systempolicy/observability.go
```diff
@@ -0,0 +1,62 @@
...
+			locSysObsProcessFileData.ProcessPaths = append(locSysObsProcessFileData.ProcessPaths, fsset...)
+		}
+		if wpfs.SetType == "File" {
```
nyrahul:
```suggestion
		if wpfs.SetType == SYS_OP_FILE {
```

src/systempolicy/observability.go
```diff
@@ -0,0 +1,62 @@
...
+	wpfs.Labels = labels
+
+	res, policyNames, err := libs.GetWorkloadProcessFileSet(CfgDB, wpfs)
+	if err != nil {
+		return err
+	}
+
+	sysObsProcessFileData := convertWPFSToObservabilityData(res, policyNames)
+	if len(sysObsProcessFileData) == 0 {
+		return errors.New("no data in db")
```
nyrahul: This should not be an error. Ok to have no data for the given filters in the DB.

nyrahul (changes requested):
please fix the new lint issues.

scripts/get_sys_observability_data.sh
seswarrajan: Yes Rahul. If we did not provide cluster name, all clusters will be read. If any vallue is ignored then * is used in query

nyrahul (approved):

PR #389, opened 2022-02-01
Modified observe script

seswarrajan:


scripts/observe_sys_data.sh
```diff
@@ -0,0 +1,28 @@
...
+        --containername ) CONTAINER_NAME="$2"; shift 2;;
+        --clustername ) CLUSTER_NAME="$2"; shift 2;;
+        --namespace ) NAMESPACE="$2"; shift 2;;
+        --labels ) LABELS="$2"; shift 2;;
+        -- ) shift; break ;;
+        * ) break ;;
+    esac
+done
+
+[[ "$CLUSTER_NAME" == "" ]] && [[ "$CONTAINER_NAME" == "" ]] && [[ "$NAMESPACE" == "" ]] && [[ "$LABELS" == "" ]] && usage
```
nyrahul: why we are printing usage() when no input is provided?

When no input is provided, we should be showing users all the DB entries.

nyrahul (changes requested):

nyrahul (approved):

PR #398, opened 2022-02-06
Clear observability DB

seswarrajan:


scripts/observe_db_clear.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
nyrahul, line 1: Let's use the existing script db_clear.sh for this purpose.
db_clear.sh is also copied in the knoxautopolicy dockerfile.

scripts/observe_db_clear.sh
```diff
@@ -0,0 +1,21 @@
...
+        --namespace ) NAMESPACE="$2"; shift 2;;
+        --labels ) LABELS="$2"; shift 2;;
+        -- ) shift; break ;;
+        * ) break ;;
+    esac
+done
+
+DATA='{"clusterName": "'$CLUSTER_NAME'", "namespace":"'$NAMESPACE'", "containerName":"'$CONTAINER_NAME'", "labels":"'$LABELS'"}'
+
+grpcurl -plaintext -d "$DATA" localhost:9089 v1.observability.Observability.ClearDb
```
nyrahul, line 21: we have an existing db-clear endpoint.. lets extend the same.

nyrahul (changes requested):

seswarrajan:
Will handle in a fresh new PR

PR #403, opened 2022-02-11
Clear WPFS DB based on duration (hr, min, sec)

seswarrajan:

### DIFF
--- a/src/systempolicy/observability.go
+++ b/src/systempolicy/observability.go
@@ -1,6 +1,9 @@
 package systempolicy
 
 import (
+	"strconv"
+	"strings"
+
 	"github.com/accuknox/auto-policy-discovery/src/libs"
 	opb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/observability"
 	types "github.com/accuknox/auto-policy-discovery/src/types"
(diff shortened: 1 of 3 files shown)


seswarrajan:
To clear the WPFS DB based on duration, the script to be used in following format. 

```
./observe_clear_sys_data.sh --req dbclear --duration xxh
./observe_clear_sys_data.sh --req dbclear --duration xxm
./observe_clear_sys_data.sh --req dbclear --duration xxs

or 

./observe_clear_sys_data.sh --req dbclear --duration xxhxxmxxs
```

src/systempolicy/observability.go
```diff
@@ -103,7 +106,44 @@ func GetSystemObsData(wpfs types.WorkloadProcessFileSet) (opb.Response, error) {
 	return opbSysObsResponse, nil
 }
 
-func ClearSysDb(wpfs types.WorkloadProcessFileSet, duration int64) error {
-	err := libs.ClearWPFSDb(CfgDB, wpfs, duration)
+func convertDurationStrToInt(durationStr string) int {
```
nyrahul: we should not write the API ourself.

Please check https://golangrepo.com/repo/xhit-go-str2duration-go-date-time

nyrahul (changes requested):

src/systempolicy/observability.go
```diff
@@ -103,7 +106,11 @@ func GetSystemObsData(wpfs types.WorkloadProcessFileSet) (opb.Response, error) {
 	return opbSysObsResponse, nil
 }
 
-func ClearSysDb(wpfs types.WorkloadProcessFileSet, duration int64) error {
-	err := libs.ClearWPFSDb(CfgDB, wpfs, duration)
+func ClearSysDb(wpfs types.WorkloadProcessFileSet, durationStr string) error {
+	if durationStr == "0" {
+		return errors.New("not a valid duration")
+	}
+	duration, _ := time.ParseDuration(durationStr)
```
nyrahul: Please handle this error.
```suggestion
	duration, err := time.ParseDuration(durationStr)
	if err != nil {
	     return err
	}
```

nyrahul (changes requested):

src/systempolicy/observability.go
```diff
@@ -103,7 +106,14 @@ func GetSystemObsData(wpfs types.WorkloadProcessFileSet) (opb.Response, error) {
 	return opbSysObsResponse, nil
 }
 
-func ClearSysDb(wpfs types.WorkloadProcessFileSet, duration int64) error {
-	err := libs.ClearWPFSDb(CfgDB, wpfs, duration)
+func ClearSysDb(wpfs types.WorkloadProcessFileSet, durationStr string) error {
+	if durationStr == "0" {
+		return errors.New("not a valid duration")
+	}
```
nyrahul, line 112: What is the point of this check? 
If the format is invalid then time.ParseDuration() will return an err.
```suggestion
```

nyrahul (changes requested):

src/systempolicy/observability.go
seswarrajan: Yes Rahul. But when we input 0, this is considered as a valid input and the output is given as 0 seconds.
That is why I explicitly included this check.

PR #410, opened 2022-02-20
Convert wpfs to systempolicy and update in sys-policy DB

seswarrajan:

### DIFF
--- a/src/systempolicy/systemPolicy.go
+++ b/src/systempolicy/systemPolicy.go
@@ -23,10 +23,7 @@ import (
 	"github.com/robfig/cron"
 )
 
-var (
-	log             *zerolog.Logger
-	isWpfsDbUpdated bool
-)
+var log *zerolog.Logger
 
 func init() {
 	log = logger.GetInstance()
@@ -87,7 +84,6 @@ var FileFromSource bool
 func init() {
 	SystemWorkerStatus = STATUS_IDLE
 	SystemStopChan = make(chan struct{})
-	isWpfsDbUpdated = false
 }
 
 // ====================== //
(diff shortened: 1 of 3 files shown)


src/systempolicy/systemPolicy.go
```diff
@@ -920,6 +956,11 @@ func PopulateSystemPoliciesFromSystemLogs(sysLogs []types.KnoxSystemLog) []types
 					polCnt = len(discoveredSysPolicies)
 					log.Info().Msgf("discovered %d file policies from %d file logs",
 						len(discoveredSysPolicies), len(fileOpLogs))
+				} else {
+					// New mode of system policy generation using WPFS table
+					if isWpfsDbUpdated {
```
nyrahul: Lets have single isWpfsDbUpdated check across file/process/network and then do updateSysPolicies() in it.

src/systempolicy/systemPolicy.go
```diff
@@ -84,6 +87,7 @@ var FileFromSource bool
 func init() {
 	SystemWorkerStatus = STATUS_IDLE
 	SystemStopChan = make(chan struct{})
+	isWpfsDbUpdated = false
```
nyrahul: Lets not keep a global flag for this purpose.

nyrahul (changes requested):

src/systempolicy/systemPolicy.go
```diff
@@ -1146,6 +1186,8 @@ func GenFileSetForAllPodsInCluster(clusterName string, pods []types.Pod, settype
 			log.Error().Msgf("failure add/updt db entry for wpfs=%+v err=%s", wpfs, err.Error())
 		}
 	}
+
+	return true
```
nyrahul: Hey, the return true should not be done blindly. Only if the DB is changed then the return should be true.

nyrahul (changes requested):

nyrahul (approved):

Issue #412, opened 2022-02-22
stable/dev image release handling 


nyrahul:
- [x] create a image in context to the branch name `accuknox/knoxautopolicy:dev` or `accuknox/knoxautopolicy:v0.2`
   - [x] on new branch create
   - [x] on push to existing branch
- [x] push stable tag to one of the v0.x release. (same as kubearmor policy)
- [x] change [default deployment](https://github.com/accuknox/auto-policy-discovery/blob/dev/deployments/k8s/deployment.yaml) to use stable tag

nyrahul:
Handled all the tasks. Closing

PR #421, opened 2022-02-24
handle stable/dev image release creation

Ankurk99:
Ref: #412


k3s/README.md
```diff
@@ -0,0 +1,8 @@
+# K3s Installation
+
+Instead of self-managed Kubernetes, you can set up K3s by running the following command.
+
+```text
+$ cd KubeArmor/contribution/k3s
```
nyrahul: Reference to kubearmor in auto-policy-discovery??

k3s/install_k3s.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2021 Authors of KubeArmor
```
nyrahul:
```suggestion
# Copyright 2022 Authors of auto-policy-discovery
```

k3s/install_k3s.sh
```diff
@@ -0,0 +1,32 @@
...
+if [ -f $KUBECONFIG ]; then
+	KUBECONFIGBKP=$KUBEDIR/config.backup
+	echo "Found $KUBECONFIG already in place ... backing it up to $KUBECONFIGBKP"
+	cp $KUBECONFIG $KUBECONFIGBKP
+fi
+
+cp /etc/rancher/k3s/k3s.yaml $KUBEDIR/config 
+
+echo "wait for initialization"
+sleep 15
```
nyrahul: Can we remove the sleep and add something [like this](https://github.com/kubearmor/KubeArmor/blob/eb71e99aebc4e1df2b01ad11f62fd0a82b5933f7/.github/workflows/latest-release.yml#L46) to check if the pods are ready?

nyrahul (changes requested):

k3s/README.md
Ankurk99: Oops, I copied the k8s setup from KubeArmor, will change it

k3s/install_k3s.sh
Ankurk99: It is waiting for k3s to start. I think we can do without it as the commands are run step-wise.

nyrahul (approved):

PR #434, opened 2022-03-14
Basic support for VM policy discovery

seswarrajan:

### DIFF
--- a/src/systempolicy/systemPolicy.go
+++ b/src/systempolicy/systemPolicy.go
@@ -716,11 +716,13 @@ func ConvertWPFSToKnoxSysPolicy(wpfsSet types.ResourceSetMap, pnMap types.Policy
 		policy.Metadata["labels"] = wpfs.Labels
 		policy.Metadata["name"] = pnMap[wpfs]
 
-		labels := strings.Split(wpfs.Labels, ",")
-		for _, label := range labels {
-			k := strings.Split(label, "=")[0]
-			v := strings.Split(label, "=")[1]
-			policy.Spec.Selector.MatchLabels[k] = v
+		if wpfs.Labels != "" {
+			labels := strings.Split(wpfs.Labels, ",")
+			for _, label := range labels {
+				k := strings.Split(label, "=")[0]
+				v := strings.Split(label, "=")[1]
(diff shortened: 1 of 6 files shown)


src/systempolicy/systemPolicy.go
```diff
@@ -936,6 +938,17 @@ func PopulateSystemPoliciesFromSystemLogs(sysLogs []types.KnoxSystemLog) []types
...
 		pods := cluster.GetPods(clusterName)
 
+		// check for vm type and if exist add default to podname
+		for _, sysLog := range sysLogs {
+			if sysLog.Namespace == types.PolicyDiscoveryHost {
+				pods = append(pods, types.Pod{
+					Namespace: types.PolicyDiscoveryHost,
+					PodName:   "default",
+				})
+				break
```
nyrahul, line 948:
```suggestion
```
Applying a break here would result in only the first syslog converted!!

src/systempolicy/systemPolicy.go
```diff
@@ -936,6 +938,17 @@ func PopulateSystemPoliciesFromSystemLogs(sysLogs []types.KnoxSystemLog) []types
 		// get k8s pods
 		pods := cluster.GetPods(clusterName)
 
+		// check for vm type and if exist add default to podname
```
nyrahul, line 941: Lets keep this loop before/outside of current for loop. Operate directly on `clusteredLogs`.

nyrahul (changes requested):

src/systempolicy/systemPolicy.go
seswarrajan: Not actually @nyrahul .
This break exits only from the inner for loop. 

When this if matches, it exits only from `for _, sysLog := range sysLogs {`

src/systempolicy/systemPolicy.go
seswarrajan: As per my understanding, we needed this loop here. Consider in case of VM, the pods will be nil. If pods are nil, clustering of logs on pod based will not work. Hence this loop will append `default` to the pods  and have the pods non nil.

PR #437, opened 2022-03-16
systempolicy: fixes for process handling

nyrahul:
With the [parent]processName fixes in kubearmor the
auto-policy-discovery is now working fine i.e. generating process names
with proper absolute system paths.

Signed-off-by: Rahul Jadhav <r@accuknox.com>

### DIFF
--- a/src/systempolicy/systemPolicy.go
+++ b/src/systempolicy/systemPolicy.go
@@ -222,8 +222,10 @@ func getSystemLogs() []types.KnoxSystemLog {
 
 		// convert kubearmor relay logs -> knox system logs
 		for _, relayLog := range relayLogs {
-			log := plugin.ConvertKubeArmorLogToKnoxSystemLog(relayLog)
-			systemLogs = append(systemLogs, log)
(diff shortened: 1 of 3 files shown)


src/systempolicy/systemPolicy.go
```diff
@@ -952,6 +954,7 @@ func PopulateSystemPoliciesFromSystemLogs(sysLogs []types.KnoxSystemLog) []types
 			}
 
 			polCnt := 0
+			isWpfsDbUpdated = false
```
seswarrajan: I suggest we can remove `isWpfsUpdated := false` from line 923

seswarrajan (changes requested):

src/systempolicy/systemPolicy.go
nyrahul: This was a bug I fixed. Earlier if you have logs from multiple clusters .. this flag was not getting reset for every cluster.. Thus the policies were regenerated regardless of any updates to the DB.

nyrahul: Fixed.

seswarrajan (approved):

PR #445, opened 2022-03-28
update go modules

Ankurk99:


nyrahul:
Hey @Ankurk99 , the CI is failing. Can you pls check and fix?

Ankurk99:
The issue is due to a field not defined in older version and is fixed in go 1.17 (ref: https://github.com/kubernetes/kubernetes/issues/106666). But max. go version supported in GA is 1.16 (ref: https://github.com/Ankurk99/auto-policy-discovery/runs/5735314020?check_suite_focus=true#step:4:33).

PR #457, opened 2022-04-17
Added sqlite3 driver

ErenYeager713:


src/libs/common.go
```diff
@@ -99,7 +99,7 @@ func SetDefaultConfig() {
 	viper.SetDefault("application.cluster.cluster-info-from", "k8sclient")
 
 	// Database config
-	viper.SetDefault("database.driver", "mysql")
+	viper.SetDefault("database.driver", "sqlite333")
```
nyrahul: what is `sqlite333`?

src/libs/dbHandler.go
```diff
@@ -26,6 +26,12 @@ func GetNetworkPolicies(cfg types.ConfigDB, cluster, namespace, status string) [
 			return results
 		}
 		results = docs
+	} else if cfg.DBDriver == "sqlite3" {
```
nyrahul, line 29: Lets just use `sqlite` as the driver name.

src/libs/dbHandler_test.go
```diff
@@ -20,7 +20,7 @@ const Unmet = "unmet expectation error: "
 // ==================== //
 
 func TestGetNetworkPolicies(t *testing.T) {
-	// prepare mock mysql
+	// prepare mock sqlite
```
nyrahul, line 103: dont change existing testcase of mysql since we will support mysql as well. Add new test cases for sqlite.

src/networkpolicy/helperFunctions.go
```diff
@@ -528,26 +528,27 @@ func updateDNSFlows(networkLogs []types.KnoxNetworkLog) {
...
 		if log.DNSRes != "" && log.DNSResIPs != nil {
 			domainName := log.DNSRes
+			newDNSIPs := log.DNSResIPs
 
 			// udpate DNS to IPs map
-			if ips, ok := DomainToIPs[domainName]; ok {
-				for _, ip := range ips {
-					if !libs.ContainsElement(ips, ip) {
-						ips = append(ips, ip)
+			if dnsIps, ok := DomainToIPs[domainName]; ok {
```
nyrahul: Why are these changes part of this PR??

src/plugin/cilium.go
```diff
@@ -253,6 +253,97 @@ func ConvertCiliumFlowToKnoxNetworkLog(ciliumFlow *cilium.Flow) (types.KnoxNetwo
 	return log, true
 }
 
+func ConvertSQLiteCiliumLogsToKnoxNetworkLogs(docs []map[string]interface{}) []types.KnoxNetworkLog {
```
nyrahul, line 256: We do not use this function anywhere I believe. Can we remove this function?

src/plugin/kubearmor.go
```diff
@@ -50,6 +50,61 @@ func ConvertKnoxSystemPolicyToKubeArmorPolicy(knoxPolicies []types.KnoxSystemPol
 	return results
 }
 
+func ConvertSQLiteKubeArmorLogsToKnoxSystemLogs(docs []map[string]interface{}) []types.KnoxSystemLog {
```
nyrahul, line 53: This function is not used as well from anywhere.

src/protobuf/v1/worker/worker.proto
```diff
@@ -18,6 +18,7 @@ message WorkerRequest {
     string namespace = 4;
     string clustername = 5;
     string labels = 6;
+    string fromsource = 7;
```
nyrahul: This change should not be part of this PR. Please rebase the code.

nyrahul (changes requested):
Thanks. Mostly looks good, few comments in the PR.

src/conf/local-file.yaml
```diff
@@ -26,7 +26,7 @@ application:
     cluster-mgmt-url: "http://localhost:8080"
 
 database:
-  driver: mysql
+  driver: sqlite3
```
seswarrajan, line 29: @ErenYeager713 
Kindly revert the change back to mysql. Once after validation we can make this change on need basis. 
Also we need to decide on the persistent directory that we will be mounting the DB to in case of K8S.

src/conf/local.yaml
```diff
@@ -24,7 +24,7 @@ application:
     cluster-mgmt-url: "http://localhost:8080"
 
 database:
-  driver: mysql
+  driver: sqlite3
```
seswarrajan, line 27: @ErenYeager713
Kindly revert the change back to mysql. Once after validation we can make this change on need basis.
Also we need to decide on the persistent directory that we will be mounting the DB to in case of K8S.

src/config/configManager_test.go
```diff
@@ -49,7 +49,7 @@ feed-consumer:
       buffer: 50
 
 database:
-  driver: mysql
+  driver: sqlite3
```
seswarrajan, line 52: @ErenYeager713
Kindly revert the change back to mysql. Once after validation we can make this change on need basis.
Also we need to decide on the persistent directory that we will be mounting the DB to in case of K8S.

src/libs/sqliteHandler.go
```diff
@@ -0,0 +1,748 @@
...
+
+func connectSQLite(cfg types.ConfigDB) (db *sql.DB) {
+	if MockDB != nil {
+		return MockDB
+	}
+
+	dbconn := cfg.DBUser + ":" + cfg.DBPass + "@tcp(" + cfg.DBHost + ":" + cfg.DBPort + ")/" + cfg.DBName
+	fmt.Println(dbconn)
+	//db, err := sql.Open(cfg.DBDriver, dbconn)
+	db, err := sql.Open(cfg.DBDriver, "/home/vagrant/accuknox.db")
```
seswarrajan: @ErenYeager713 Please do not hard-code any paths. Get this path from local-file config and use the same across the code.

src/libs/sqliteHandler.go
```diff
@@ -0,0 +1,748 @@
...
+}
+
+func connectSQLite(cfg types.ConfigDB) (db *sql.DB) {
+	if MockDB != nil {
+		return MockDB
+	}
+
+	dbconn := cfg.DBUser + ":" + cfg.DBPass + "@tcp(" + cfg.DBHost + ":" + cfg.DBPort + ")/" + cfg.DBName
+	fmt.Println(dbconn)
+	//db, err := sql.Open(cfg.DBDriver, dbconn)
```
seswarrajan: @ErenYeager713  Remove this commented line.

seswarrajan (changes requested):
Please rebase the code and update the PR.
Also address the review comments

src/plugin/kubearmor.go
ErenYeager713: This function is used here https://github.com/ErenYeager713/auto-policy-discovery/blob/dev/src/systempolicy/systemPolicy.go

src/protobuf/v1/worker/worker.proto
ErenYeager713: Okay

src/networkpolicy/helperFunctions.go
ErenYeager713: WIll rebase the code again

src/conf/local-file.yaml
ErenYeager713: Sure

src/libs/sqliteHandler.go
ErenYeager713: Okay

src/libs/dbHandler.go
ErenYeager713: If we don't specify `sqlite3` then it doesn't identify the sqlite driver

seswarrajan: The driver name should match the name as per the driver value in local-file config file. So sqlite should work.

src/plugin/cilium.go
ErenYeager713: It is used in cilium.go file itself
```
func ConvertCiliumNetworkLogsToKnoxNetworkLogs(dbDriver string, docs []map[string]interface{}) []types.KnoxNetworkLog {
	if dbDriver == "mysql" {
		return ConvertMySQLCiliumLogsToKnoxNetworkLogs(docs)
	} else if dbDriver == "sqlite3" {
                return ConvertSQLiteCiliumLogsToKnoxNetworkLogs(docs)
        } else if dbDriver == "mongo" {
		return ConvertMongodCiliumLogsToKnoxNetworkLogs(docs)
	} else {
		return []types.KnoxNetworkLog{}
	}
}
```

src/libs/dbHandler.go
ErenYeager713: Yeah so I tried with `sqlite` and it doesn't recognise the driver, also in the local-file the name of the driver is `sqlite3` , because anything else and it doesn't recognise the driver

src/systempolicy/systemPolicy.go
```diff
@@ -1279,7 +1279,7 @@ func GenFileSetForAllPodsInCluster(clusterName string, pods []types.Pod, settype
 		} else {
 			if !reflect.DeepEqual(mergedfs, out[wpfs]) {
 				log.Info().Msgf("updating wpfs db entry for wpfs=%+v", wpfs)
-				err = libs.UpdateWorkloadProcessFileSetMySQL(CfgDB, wpfs, mergedfs)
+				err = libs.UpdateWorkloadProcessFileSetSQLite(CfgDB, wpfs, mergedfs)
```
nyrahul, line 1282: This will not work when the driver is set to mysql. Note that we also have to support mysql.
CC @seswarrajan

nyrahul (changes requested):
Note that we are not deprecating mysql. We are adding one more DB driver in the form of sqlite. Both mysql and sqlite needs to work.
With the current changes, mysql will stop working.

PR #458, opened 2022-04-18
Added support for network insight

seswarrajan:


src/insight/networkInsight.go
```diff
@@ -0,0 +1,269 @@
...
+		} else {
+			networkData = append(networkData, ipb.NetworkInsightData{
+				ClusterName: locRes.ClusterName,
+				Namespace:   locRes.Namespace,
+				Labels:      locRes.Labels,
+				Type:        locRes.Type,
+				Rule:        locRes.Rule,
+				NetResource: locRes.NetResource,
+			})
+		}
```
Ankurk99: Please simplify this duplicate code. What will be the case when `req.Labels != "" and req.Labels != locRes.Labels`?

src/insight/systemInsight.go
```diff
@@ -0,0 +1,181 @@
...
+		locFsData.FromSource = wpfs.FromSource
+		if wpfs.SetType == sys.SYS_OP_FILE {
+			locFsData.FilePaths = append(locFsData.FilePaths, fsset...)
+		}
+		if wpfs.SetType == sys.SYS_OP_PROCESS {
+			locFsData.ProcessPaths = append(locFsData.ProcessPaths, fsset...)
+		}
+		if wpfs.SetType == sys.SYS_OP_NETWORK {
+			locFsData.NetworkPaths = append(locFsData.NetworkPaths, fsset...)
+		}
```
Ankurk99, line 50: can use else-if?

src/insight/systemInsight.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		} else {
+			locObsData.ClusterName = wpfs.ClusterName
+			locObsData.Namespace = wpfs.Namespace
+			locObsData.Labels = wpfs.Labels
+			locObsData.ContainerName = wpfs.ContainerName
+			locObsData.SysProcessFileData = append(locObsData.SysProcessFileData, locFsData)
+
+			resData.SysData = append(resData.SysData, locObsData)
+		}
```
Ankurk99, line 80: duplication

src/protobuf/v1/insight/insight.proto
```diff
@@ -0,0 +1,114 @@
...
+    bool Aggregated = 3;
+}
+
+message Ingress {
+    map<string, string> MatchLabels = 1;
+    repeated SpecPort ToPorts = 2;
+    repeated SpecHTTP ToHTTPs = 3;
+    repeated SpecCIDR FromCIDRs = 4;
+    repeated string FromEntities = 5;
+}
```
Ankurk99: add an EOF line

src/insight/systemInsight.go
seswarrajan: No... We cannot use else-if here as all 3 cases are possible in same scenario.

src/insight/systemInsight.go
seswarrajan: I will check if it is possible to ignore if else and bring in one if case.

seswarrajan: This duplication cannot be ignored.
If case in non-zero case and else case runs only once when index = 0. 

I will check the possibility to bring all under one loop.

Ankurk99:
I think you would also need to make changes to kubearmor-client to use insight protobuf instead of observability protobuf (for eg here: https://github.com/kubearmor/kubearmor-client/blob/main/insight/insight.go#L14)

Ankurk99:
> I think you would also need to make changes to kubearmor-client to use insight protobuf instead of observability protobuf (for eg here: [https://github.com/kubearmor/kubearmor-client/blob/main/insight/insight.go#L14](https://github.com/kubearmor/kubearmor-client/blob/main/insight/insight.go?rgh-link-date=2022-04-18T14%3A37%3A30Z#L14))

As discussed in our chat, these changes will be made after this PR gets merged as karmor dependability should be a different issue.

Ankurk99:
LGTM

Ankurk99 (approved):

Ankurk99 (approved):

nyrahul (approved):

PR #460, opened 2022-04-19
added sqlite driver

ErenYeager713:
Signed-off-by: ErenYeager713 <gourav@accuknox.com>

### DIFF
--- a/src/conf/local-file.yaml
+++ b/src/conf/local-file.yaml
@@ -32,6 +32,7 @@ database:
   user: root
   password: password
   dbname: accuknox
+  sqlite-db-path: ./accuknox.db
   table-network-log: network_log
   table-network-policy: network_policy
   table-system-log: system_log
--- a/src/conf/local.yaml
+++ b/src/conf/local.yaml
@@ -30,6 +30,7 @@ database:
   user: root
   password: password
   dbname: accuknox
+  sqlite-db-path: ./accuknox.db
   table-network-log: network_log
   table-network-policy: network_policy
   table-system-log: system_log
--- a/src/libs/common.go
+++ b/src/libs/common.go
@@ -104,6 +104,7 @@ func SetDefaultConfig() {
 	viper.SetDefault("database.dbname", "accuknox")
 	viper.SetDefault("database.host", "127.0.0.1")
 	viper.SetDefault("database.port", "3306")
+	viper.SetDefault("database.sqlite-db-path","./accuknox.db")
 	viper.SetDefault("database.table-network-policy", "network_policy")
 	viper.SetDefault("database.table-system-policy", "system_policy")
 
--- a/src/libs/sqliteHandler.go
+++ b/src/libs/sqliteHandler.go
@@ -1,0 +1,745 @@
+package libs
+
+import (
+	"database/sql"
+	"encoding/json"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/accuknox/auto-policy-discovery/src/types"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+const WorkloadProcessFileSetSQLite_TableName = "workload_process_fileset"
+const TableNetworkPolicySQLite_TableName = "network_policy"
+const TableSystemPolicySQLite_TableName = "system_policy"
+
+// ================ //
+// == Connection == //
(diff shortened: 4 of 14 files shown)


src/conf/local-file.yaml
```diff
@@ -32,6 +32,7 @@ database:
   user: root
   password: password
   dbname: accuknox
+  sqlite-db-path: /home/vagrant/accuknox.db
```
nyrahul: This cannot be used as the path since the /home/vagrant path might not be available on the target system. Lets use current directory for keeping accuknox.db

src/conf/local.yaml
```diff
@@ -30,6 +30,7 @@ database:
   user: root
   password: password
   dbname: accuknox
+  sqlite-db-path: /home/vagrant/accuknox.db
```
nyrahul: Same comment as before.

src/libs/common.go
```diff
@@ -104,6 +104,7 @@ func SetDefaultConfig() {
 	viper.SetDefault("database.dbname", "accuknox")
 	viper.SetDefault("database.host", "127.0.0.1")
 	viper.SetDefault("database.port", "3306")
+	viper.SetDefault("database.sqlite-dp-path","/home/vagrant/accuknox.db")
```
nyrahul: change the default path to current directory.

nyrahul (changes requested):

src/libs/common.go
```diff
@@ -104,6 +104,7 @@ func SetDefaultConfig() {
 	viper.SetDefault("database.dbname", "accuknox")
 	viper.SetDefault("database.host", "127.0.0.1")
 	viper.SetDefault("database.port", "3306")
+	viper.SetDefault("database.sqlite-dp-path","./accuknox.db")
```
Ankurk99: "database.sqlite-db-path"

src/libs/sqliteHandler.go
```diff
@@ -0,0 +1,745 @@
...
+
+func GetNetworkPoliciesFromSQLite(cfg types.ConfigDB, cluster, namespace, status string) ([]types.KnoxNetworkPolicy, error) {
+	db := connectSQLite(cfg)
+	defer db.Close()
+
+	policies := []types.KnoxNetworkPolicy{}
+	var results *sql.Rows
+	var err error
+
+	query := "SELECT apiVersion,kind,flow_ids,name,cluster_name,namespace,type,rule,status,outdated,spec,generatedTime FROM " + TableNetworkPolicySQLite_TableName
```
Ankurk99, line 84: cc @nyrahul Should we be concerned of SQL injection and create prepared statements objects (using sqlite3_prepare()) for the queries?

src/libs/sqliteHandler.go
```diff
@@ -0,0 +1,745 @@
...
+		results, err = db.Query(query, cluster, status)
+	} else if namespace != "" && status != "" {
+		query = query + " WHERE namespace = ? and status = ? "
+		results, err = db.Query(query, namespace, status)
+	} else if status != "" {
+		query = query + " WHERE status = ? "
+		results, err = db.Query(query, status)
+	} else {
+		results, err = db.Query(query)
+	}
```
Ankurk99, line 99: Can you please confirm if all the conditions are satisfied here? (cluster and namespace condition). Also can we simplify this logic? it's a bit difficult to read

src/libs/sqliteHandler.go
nyrahul: This particular statement won't result in SQLI ... go-sec automatically identifies scenarios where SQLI could be used ... Typically SQLI is used where the query is constructed from non-constant variables. In this case, the query is prepared from constant strings + constant var only and hence no risk.

nyrahul (approved):

nyrahul:
The code  changes look good to me. Can you please check CI failures? Thanks

ErenYeager713:
> The code changes look good to me. Can you please check CI failures? Thanks

Yeah

src/libs/sqliteHandler.go
ErenYeager713: Sure @Ankurk99 , will try to simplify it

seswarrajan (approved):
LGTM

seswarrajan (approved):
LGTM

nyrahul:
Great work @ErenYeager713

PR #465, opened 2022-05-19
Set source to "" only on process operation

seswarrajan:


src/plugin/kubearmor.go
```diff
@@ -201,11 +201,11 @@ func ConvertKubeArmorLogToKnoxSystemLog(relayLog *pb.Log) (types.KnoxSystemLog,
 		readOnly = true
 	}
 
-	if strings.Contains(source, "runc") {
+	if strings.Contains(source, "runc") && relayLog.Operation == "Process" {
```
nyrahul, line 204: There is a global constant to identify "Process". Lets use that.

nyrahul (commented):
@seswarrajan , even without this PR the process based rules should be added.  The only check that is added is to additionally check "Process" as the condition. I believe we do not need this PR.

PR #470, opened 2022-05-25
Modifying json import to support ignoring empty json values

seswarrajan:

### DIFF
--- a/src/libs/common.go
+++ b/src/libs/common.go
@@ -458,7 +458,10 @@ func WriteCiliumPolicyToYamlFile(namespace string, policies []types.CiliumNetwor
 	}
 
 	for i := range policies {
-		jsonBytes := json.Marshal(&policies[i])
+		jsonBytes, err := json.Marshal(&policies[i])
+		if err != nil {
+			log.Error().Msg(err.Error())
+		}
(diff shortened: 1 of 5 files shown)


src/libs/common.go
```diff
@@ -458,7 +458,10 @@ func WriteCiliumPolicyToYamlFile(namespace string, policies []types.CiliumNetwor
 	}
 
 	for i := range policies {
-		jsonBytes := json.Marshal(&policies[i])
+		jsonBytes, err := json.Marshal(&policies[i])
+		if err != nil {
+			log.Error().Msg(err.Error())
```
nyrahul, line 463: If json marshalling fails then we can't subsequently use the `jsonBytes`. In the error condition we should either `continue` or `break` as appropriate.

src/libs/common.go
```diff
@@ -488,7 +491,10 @@ func WriteKubeArmorPolicyToYamlFile(fname string, policies []types.KubeArmorPoli
...
```
nyrahul, line 498: Same as comment above. Also handle the same for JSONToYAML API used below.

nyrahul (changes requested):

nyrahul (approved):
LGTM

PR #482, opened 2022-06-15
Code cleanup and fix for creating policy under dir mentioned in config

seswarrajan:


src/libs/common.go
```diff
@@ -438,7 +441,16 @@ func WriteKnoxNetPolicyToYamlFile(namespace string, policies []types.KnoxNetwork
...
-	fileName := GetEnv("POLICY_DIR", "./")
+
+	policyDir := cfg.CurrentCfg.ConfigNetPolicy.NetworkPolicyDir
+	var fileName string
+
+	if policyDir != "./" {
+		fileName = GetEnv("POLICY_DIR", "./")
+	} else {
+		fileName = policyDir
+	}
```
nyrahul, line 452: This code does not make sense.
If the NetworkPolicyDir in the configmap is set to any value other thant "./" then it will not be used!!

src/libs/common.go
```diff
@@ -477,7 +489,16 @@ func WriteCiliumPolicyToYamlFile(namespace string, policies []types.CiliumNetwor
...
+
+	policyDir := cfg.CurrentCfg.ConfigSysPolicy.SystemPolicyDir
+	var fileName string
+
+	if policyDir != "./" {
+		fileName = GetEnv("POLICY_DIR", "./")
+	} else {
+		fileName = policyDir
+	}
+
```
nyrahul, line 501: This code does not make sense.
If the SystemPolicyDir in the configmap is set to any value other thant "./" then it will not be used!!

src/libs/common.go
```diff
@@ -565,3 +586,50 @@ func ConvertStrToUnixTime(strTime string) int64 {
...
+		}
+	}
+	return str
+
+}
+
+//ConvertStringToArray - Convert String to Array of string
+func ConvertStringToArray(str string) []string {
+	return strings.Split(str, ",")
+}
```
nyrahul, line 611: Why is this api needed if its just a single line of code within it?

src/libs/common.go
```diff
@@ -565,3 +586,50 @@ func ConvertStrToUnixTime(strTime string) int64 {
...
+
+func ConvertFilterString(filter []string) string {
+	var query string
+	//Create the filter query
+	for i, value := range filter {
+		query = query + constants.BACK_SLASH + fmt.Sprint(value) + constants.BACK_SLASH
```
nyrahul, line 617: I think having constants define is not helping here! It was better to use literals directly.
```suggestion
		query = query + "\" + fmt.Sprint(value) + "\"
```

nyrahul (changes requested):

src/libs/common.go
seswarrajan: This is just an existing implementation porting from observability to discovery-engine. 
Will handle this .

src/libs/common.go
seswarrajan: This is just an existing implementation porting from observability to discovery-engine. 
Will handle this .

src/libs/common.go
seswarrajan: No Rahul. As per implementation, if it is any value other than **./**, then the same will be used.

src/libs/common.go
seswarrajan: No Rahul. As per implementation, if it is any value other than **./**, then the same will be used.

seswarrajan:
Handling both changes separately in a diff PR. Hence closing the same.

PR #483, opened 2022-06-16
Fix for nw/sys policy dir path read from config

seswarrajan:

### DIFF
--- a/src/libs/common.go
+++ b/src/libs/common.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/clarketm/json"
 
+	cfg "github.com/accuknox/auto-policy-discovery/src/config"
 	logger "github.com/accuknox/auto-policy-discovery/src/logging"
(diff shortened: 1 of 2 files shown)


src/libs/common.go
```diff
@@ -376,6 +377,14 @@ func GetCommandOutput(cmd string, args []string) string {
 // == File I/O == //
 // ============== //
 
+func getPolicyDir(cfgPath string) string {
+	if cfgPath == "" {
+		return GetEnv("POLICY_DIR", "./")
+	} else {
+		return cfgPath
+	}
```
nyrahul:
```suggestion
	if cfgPath == "" {
		return GetEnv("POLICY_DIR", "./")
	}
	return cfgPath
```

nyrahul (changes requested):

nyrahul (approved):

PR #485, opened 2022-06-18
stable release GH action fixes

nyrahul:
Lets assume we have v0.1 branch as the stable version branch.
Now if a new code or a backport is pushed ot the v0.1 branch then the
stable image should also be updated. This was not getting updated
because the stable image push was done even before the v0.1 image was
uploaded (since the actions were operating in parallel). We now have
added the stable release update as a different job in the same action
yaml thus ensuring that it executes serially.

Signed-off-by: Rahul Jadhav <nyrahul@gmail.com>


.github/workflows/latest-release.yml
```diff
@@ -57,3 +57,38 @@ jobs:
       - name: Push discovery engine images to Docker
         run: |
             ./src/build/push_knoxautopolicy.sh ${{ steps.vars.outputs.tag }}
+
+  push-stable-version:
+    name: Create discovery engine Release - 18.04
```
Ankurk99, line 62: Please add a `needs: build` so that this job runs only after the `build` is done.

.github/workflows/latest-release.yml
```diff
@@ -57,3 +57,38 @@ jobs:
...
+    if: github.repository == 'accuknox/discovery-engine'
```
Ankurk99: We won't need this as we are already checking this in `build`, instead replace it with `if: github.ref != 'refs/heads/dev'` because this action is only for non `dev` branches, as `dev` branch is already handled in stable-release.yml.

.github/workflows/latest-release.yml
nyrahul: Perfect! Thanks for the point.

PR #492, opened 2022-06-27
Do not include IP address if IP/port empty

seswarrajan:


nyrahul:
I am not sure why we are getting AF_INET events without IP address/port? That needs to be checked.

PR #495, opened 2022-07-03
Fix for AF_UNIX/AF_INET in summary output + K8s Pod watcher code implemented

seswarrajan:


src/observability/watcher.go
```diff
@@ -0,0 +1,80 @@
...
+					Namespace: event.Object.ObjectMeta.Namespace,
+					PodName:   event.Object.ObjectMeta.Name,
+					IP:        event.Object.Status.PodIP,
+				}
+
+				if event.Type == "ADDED" {
+					addPodToList(pod)
+				} else if event.Type == "MODIFIED" {
+					updatePodList(pod)
+				}
```
Ankurk99, line 101: Should we also be deleting the pod's information from the list?

seswarrajan: Not rquired @Ankurk99 
The reason is that the sumarry logs stays in DB even when pods are removed. 
So when collecting summary output, we need to know the existing pod details as well.

Ankurk99 (approved):

src/cluster/k8sWatcher.go
```diff
@@ -0,0 +1,97 @@
...
+		resp, err := watchClient.Do(req)
+		if err != nil {
+			return nil
+		}
+
+		return resp
+	}
+
+	// kube-proxy (local)
+	URL := "http://" + host + ":" + port + "/api/v1/services?watch=true"
```
nyrahul, line 89:
```suggestion
	URL := "https://" + host + ":" + port + "/api/v1/services?watch=true"
```

src/cluster/k8sWatcher.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// #nosec
+	if resp, err := http.Get(URL); err == nil {
+		return resp
```
nyrahul, line 93: print error

src/cluster/k8sWatcher.go
```diff
@@ -0,0 +1,97 @@
...
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			},
+		}
+
+		URL := "https://" + host + ":" + port + "/api/v1/services?watch=true"
+
+		req, err := http.NewRequest("GET", URL, nil)
+		if err != nil {
+			return nil
```
nyrahul, line 74: print error statement in all failure cases, since we are not returning err

src/observability/summary.go
```diff
@@ -23,46 +23,77 @@ func deDuplicateServerInOutConn(connList []types.SysNwConnDetail) []types.SysNwC
...
-		resslice := strings.Split(res, " ")
+	if strings.Contains(log.Resource, "AF_INET") {
+		var ip, port string
+		resslice := strings.Split(log.Resource, " ")
 		for _, locres := range resslice {
 			if strings.Contains(locres, "sin_addr") {
-				conn.Path = strings.Split(locres, "=")[1]
+				ip = strings.Split(locres, "=")[1]
 			}
 			if strings.Contains(locres, "sin_port") {
```
nyrahul, line 63:
```suggestion
			else if strings.Contains(locres, "sin_port") {
```

nyrahul (changes requested):

seswarrajan:
performing different handling. Will be opening a new PR.

PR #498, opened 2022-07-07
resource-update

sanjay-ba:


deployments/helm/values-dev.yaml
```diff
@@ -80,11 +80,11 @@ ingress:
...
   requests:
-    cpu: 800m
-    memory: 2Gi
+    cpu: 20m
+    memory: 30Mi
   limits:
-    cpu: 800m
-    memory: 2Gi
+    cpu: 40m
+    memory: 50Mi
```
wazir-ahmed, line 87: @sanjay-ba I once observed in the `saas-dev` or `saas-prod` cluster the DE is crashing (because of OOM Killer) when such low values are set (I did not remember the exact values). Have you tested whether the DE is able to generate policies with these values/limits configured?

PR #506, opened 2022-07-18
Handling/Adding preconfigured rule for Kubearmor policy

seswarrajan:

### DIFF
--- a/src/plugin/kubearmor.go
+++ b/src/plugin/kubearmor.go
@@ -267,11 +267,8 @@ func ConvertKubeArmorLogToKnoxSystemLog(relayLog *pb.Log) (types.KnoxSystemLog, 
 		resource = ""
 	}
 
-	if strings.Contains(resource, types.PreConfiguredKubearmorRule) {
-		if resource[0:len(types.PreConfiguredKubearmorRule)] == types.PreConfiguredKubearmorRule {
-			log.Info().Msgf("predefined resource [%s]. %+v", resource, relayLog)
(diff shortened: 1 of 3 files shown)


src/plugin/kubearmor.go
```diff
@@ -259,6 +267,13 @@ func ConvertKubeArmorLogToKnoxSystemLog(relayLog *pb.Log) (types.KnoxSystemLog,
 		resource = ""
 	}
 
+	if strings.Contains(resource, types.PreConfiguredKubearmorRule) {
+		if resource[0:len(types.PreConfiguredKubearmorRule)] == types.PreConfiguredKubearmorRule {
+			log.Info().Msgf("predefined resource [%s]. %+v", resource, relayLog)
+			return types.KnoxSystemLog{}, errors.New("predefined file resource")
+		}
+	}
```
daemon1024:
```suggestion
	if strings.HasPrefix(resource,types.PreConfiguredKubearmorRule) { {
		// log.Info().Msgf("predefined resource [%s]. %+v", resource, relayLog)
		return types.KnoxSystemLog{}, errors.New("predefined file resource")
	}
```

This predefined resource is used pretty often, so not sure if we should be logging it or not.

seswarrajan: That is why I did not plan to throw it as an error but an info instead. If we think the logging will be quite continuous, we can ignore then.

daemon1024: I have also suggested an improvement to the conditional, can you check that once?

daemon1024 (approved):

PR #512, opened 2022-07-23
Extract cluster IP and POD IP with cluster management APIs

seswarrajan:


src/libs/mysqlHandler.go
```diff
@@ -79,12 +79,12 @@ func GetNetworkPoliciesFromMySQL(cfg types.ConfigDB, cluster, namespace, status,
 	}
 
 	results, err = DBHandle.Query(query+whereClause, args...)
-	defer results.Close()
 
 	if err != nil {
 		log.Error().Msg(err.Error())
 		return nil, err
 	}
+	defer results.Close()
```
nyrahul, line 88: This change seems to be incorrect. The `Close()` should be done only if the `Query()` is successful. The `results.Close()` should not be done if the `err != nil`.

nyrahul (changes requested):

src/libs/dbHandler.go
```diff
@@ -1,11 +1,57 @@
...
+// WaitForDB
+func WaitForDB(db *sql.DB) {
+	for {
+		err := db.Ping()
+		if err != nil {
+			time.Sleep(time.Second * 1)
+			log.Error().Msgf("DB Ping() failed. Will retry. err=%s", err.Error())
+		} else {
+			break
+		}
```
nyrahul:
```suggestion
		err := db.Ping()
		if err == nil {
			break
		}
		time.Sleep(time.Second * 1)
		log.Error().Msgf("DB Ping() failed. Will retry. err=%s", err.Error())
```

src/libs/dbHandler.go
```diff
@@ -1,11 +1,57 @@
...
+func NewMockDB() (*sql.DB, sqlmock.Sqlmock) {
+	db, mock, err := sqlmock.New()
+	if err != nil {
+		log.Error().Msgf("an error '%s' was not expected when opening a stub database connection", err)
+	}
+
+	MockSql = mock
+	MockDB = db
+
+	return db, mock
```
nyrahul: Is there a point in returning the db pointers as well as setting the global values? This is not a good practice.

src/libs/dbHandler.go
```diff
@@ -1,11 +1,57 @@
...
+			time.Sleep(time.Second * 1)
+			log.Error().Msgf("DB Ping() failed. Will retry. err=%s", err.Error())
+		} else {
+			break
+		}
+	}
+}
+
+func NewMockDB() (*sql.DB, sqlmock.Sqlmock) {
+	db, mock, err := sqlmock.New()
```
nyrahul: Why to add this function is `dbHandler.go` when it is used only from `_test` file?

nyrahul (changes requested):

src/libs/dbHandler.go
seswarrajan: @nyrahul This function existed in both DB files. Hence maintained in the same location.

src/libs/dbHandler.go
seswarrajan: This is as part of existing code. Hence cleaning up. Will handle this.

src/libs/mysqlHandler.go
seswarrajan: Had the same confusion, but moved this handle as per existing db functions. Will handle this.

src/libs/dbHandler.go
nyrahul: But according to this commit, this is a newly added function?

src/libs/mysqlHandler.go
seswarrajan: Screenshot from 2022-07-25 16-35-04

seswarrajan: Suggestion is to check error before deferring results.Close() or stmt.Close()

src/libs/dbHandler.go
seswarrajan: Yes @nyrahul
This was earler handled in sqlitehandler and mysqlhandler.go.

Hence thought of moving to a common position dbhandler.go. 
But currently moved to dbhandler_test.go. Thanks for pointing it out.

wazir-ahmed (changes requested):
@seswarrajan 
The changes looks good to me. But I think we missed a scenario with the DB connection handler. We only connect once during `InitDB()`. What will happen if the connection is disconnected (due to some network failure)? We have to periodically `Ping()` and check if the connection is active, if not, we have to reconnect.

We also have to protect the `DBHandler` pointer with some kind of lock since it could get updated & accessed from two goroutines at the same time (The reconnect logic updating the pointer and sql functions accessing the pointer).

wazir-ahmed:
@nyrahul @seswarrajan This PR is becoming bigger day by day. It has also become a PR about many things. Please don't add any more commit on this PR. Please merge this PR if this is tested.

seswarrajan:
> 

Yes @wazir-ahmed 
I just wanted to make sure not to miss any changes, so making this a draft PR. Will try to split and raise PR for relevant topics.

PR #514, opened 2022-07-27
support observability for containerized workloads

Ankurk99:


src/systempolicy/systemPolicy.go
```diff
@@ -1132,7 +1136,10 @@ func PopulateSystemPoliciesFromSystemLogs(sysLogs []types.KnoxSystemLog) []types
 
 func GetPodLabels(cn string, pn string, ns string, pods []types.Pod) ([]string, error) {
 	for _, pod := range pods {
-		if pod.Namespace == ns && pod.PodName == pn {
+		if pod.Namespace == "accuknox-vm-namespace" && pod.PodName == "accuknox-vm-podname" {
```
seswarrajan:
```suggestion
if pod.Namespace == types.PolicyDiscoveryVMNamespace && types.PodName == types.PolicyDiscoveryVMPodName {
```

src/systempolicy/systemPolicy.go
```diff
@@ -1261,8 +1268,11 @@ func GenFileSetForAllPodsInCluster(clusterName string, pods []types.Pod, settype
 		wpfs.FromSource = slog.Source
 		wpfs.SetType = settype
 		labels, err := GetPodLabels(slog.ClusterName, slog.PodName, slog.Namespace, pods)
+		if labels == nil {
```
seswarrajan: Please explain this line of code.

seswarrajan (changes requested):

src/systempolicy/systemPolicy.go
Ankurk99: In case of un-orchestrated containers, we won't be receiving any labels, so the labels are derived from the containername

seswarrajan: Agreed. But before performing such operation please do make sure to check if the log is related to un-orchastrated containers.

src/systempolicy/systemPolicy.go
```diff
@@ -844,6 +844,10 @@ func ConvertWPFSToKnoxSysPolicy(wpfsSet types.ResourceSetMap, pnMap types.Policy
 
 func getPodInstance(key SysLogKey, pods []types.Pod) (types.Pod, error) {
 	for _, pod := range pods {
+		if key.Namespace != "" || key.PodName != "" {
+			return pod, nil
+		}
```
nyrahul: I have doubts about this check. A non-nil value does not mean that the pod matches it. Didn't understand the rational for this check. This check might impact existing behaviour in case of k8s orchestrated workloads.

src/conf/local-file.yaml
```diff
@@ -1,7 +1,7 @@
 application:
   name: knoxautopolicy
   network:
-    operation-mode: 1                             # 1: cronjob | 2: one-time-job
+    operation-mode: 0                             # 1: cronjob | 2: one-time-job
```
nyrahul: Please do not disable this by default.

nyrahul (changes requested):
Mostly LGTM. Just one change in the defaults in conf file.

src/conf/local-file.yaml
Ankurk99: Okay

nyrahul (approved):
LGTM

PR #519, opened 2022-08-05
Extract svc/pod IP from cluster info

seswarrajan:


nyrahul (approved):
LGTM 👍🏼

PR #521, opened 2022-08-08
Creating new config under configmap for observability

seswarrajan:

### DIFF
--- a/src/conf/local-file.yaml
+++ b/src/conf/local-file.yaml
@@ -30,7 +30,7 @@ application:
 observability: 
   enable: true
   cron-job-time-interval: "0h0m10s"         # format: XhYmZs
-  dbname: ./accuknox-pol.db
+  dbname: ./accuknox-obs.db
 
 database:
   driver: sqlite3
@@ -39,7 +39,7 @@ database:
   user: root
   password: password
   dbname: accuknox
-  sqlite-db-path: ./accuknox.db
+  sqlite-db-path: ./accuknox-pol.db
   table-network-log: network_log
   table-network-policy: network_policy
   table-system-log: system_log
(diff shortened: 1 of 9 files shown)


src/conf/local-file.yaml
```diff
@@ -27,7 +27,10 @@ application:
     #cluster-mgmt-url: "http://cluster-management-service.accuknox-dev-cluster-mgmt.svc.cluster.local/cm"
     cluster-mgmt-url: "http://localhost:8080"
 
-observability: true
+observability: 
+  enable: true
+  cron-job-time-interval: "0h0m10s"         # format: XhYmZs
```
nyrahul, line 32: This cron job config is never used in the code.

src/config/configManager.go
```diff
@@ -382,3 +381,19 @@ func GetCfgClusterInfoFrom() string {
...
+
+// ============================ //
+// == Get Observability Info == //
+// ============================ //
+
+func GetCfgObservabilityEnable() bool {
+	return CurrentCfg.ConfigObservability.Enable
+}
+
+func GetCfgObservabilityCronJobTime() string {
```
nyrahul, line 393: Where is this API getting used?

src/networkpolicy/networkPolicy.go
```diff
@@ -2164,8 +2164,8 @@ func StartNetworkCronJob() {
 		log.Error().Msg(err.Error())
 		return
 	}
-	if cfg.CurrentCfg.Observability {
-		err := NetworkCronJob.AddFunc(cfg.GetCfgSysCronJobTime(), observability.NetworkLogCronJob) // time interval
+	if cfg.GetCfgObservabilityEnable() {
+		err := NetworkCronJob.AddFunc(cfg.GetCfgObservabilityCronJobTime(), observability.NetworkLogCronJob) // time interval
```
nyrahul, line 2168: Why do we need a cron job for saving the network/system logs?
In the case of discovery-engine before, we used cron-job because we had to periodically discover the policies .. but in the casee of observability we do not have that requirement. 
We should have a go routine which perpetually waits on the network/system events and pushes it to the log without any cron settings.

nyrahul (changes requested):

nyrahul (approved):

PR #522, opened 2022-08-10
Observability : handle log update/insert in a single function

seswarrajan:

### DIFF
--- a/deployments/k8s/deployment.yaml
+++ b/deployments/k8s/deployment.yaml
@@ -59,7 +59,12 @@ data:
       cluster:
         cluster-info-from: "k8sclient"            # k8sclient|accuknox
 
-    observability: false
+    observability: 
+      enable: false
+      cron-job-time-interval: "0h0m5s"         # format: XhYmZs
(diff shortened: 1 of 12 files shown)


wazir-ahmed (approved):
LGTM. 

@seswarrajan But I have a question. We still open/close DB connection for each query, right? We have not changed that design even in case of auto-policy-generation, right?

seswarrajan:
> LGTM.
> 
> @seswarrajan But I have a question. We still open/close DB connection for each query, right? We have not changed that design even in case of auto-policy-generation, right?

Yes @wazir-ahmed 
I haven't changed the design yet. Having a single handle has lot of performance and design implications. Since multiple go routines handles the same DBHandle, there is a DB lock issue easily observed. 
If same DBHandle is used, sqlite has a problem in concurrency and when internal concurrency is enabled, the DB read/write operations are slowed down. 

This implementation of same DBHandle needs more time and intense testing.
Hence suggest to handle as separate user story.

wazir-ahmed:
@seswarrajan Okay, we can handle that in the future. No problem.

PR #524, opened 2022-08-16
package discovery-engine as a systemd service

Ankurk99:
Signed-off-by: Ankur Kothiwal <ankur.kothiwal@accuknox.com>

### DIFF
--- a/src/.goreleaser.yml
+++ b/src/.goreleaser.yml
@@ -1,27 +1,71 @@
-# This is an example .goreleaser.yml file with some sensible defaults.
-# Make sure to check the documentation at https://goreleaser.com
+project_name: knoxAutoPolicy
+
 before:
   hooks:
     # You may remove this if you don't use go modules.
     - go mod tidy
     # you may remove this if you don't need go generate
     - go generate ./...
+
 builds:
-  - env:
-      - CGO_ENABLED=1
+  - binary: "opt/knoxAutoPolicy/src/knoxAutoPolicy"
+    id: knoxAutoPolicy
     goos:
       - linux
         #- windows
         #- darwin
     goarch:
       - amd64
+      
 archives:
-  - replacements:
-      darwin: Darwin
-      linux: Linux
-      windows: Windows
-      386: i386
-      amd64: x86_64
+  - id: "knoxAutoPolicy"
+    builds:
+      - "knoxAutoPolicy"
+    name_template: "{{.ProjectName}}_{{.Version}}_{{.Os}}-{{.Arch}}"
+    files:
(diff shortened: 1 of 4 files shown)


src/packaging/discovery-engine.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=knoxAutoPolicy
+
+[Service]
+User=root
```
nyrahul: Do we really need root user?

nyrahul (changes requested):
Mostly LGTM, except that the `user=root` is enabled for knoxAutoPolicy.

src/packaging/discovery-engine.service
seswarrajan: We do not need this to run as super user/root user.

src/.goreleaser.yml
```diff
@@ -1,27 +1,71 @@
...
-  - env:
-      - CGO_ENABLED=1
+  - binary: "opt/knoxAutoPolicy/src/knoxAutoPolicy"
+    id: knoxAutoPolicy
     goos:
       - linux
         #- windows
         #- darwin
     goarch:
       - amd64
```
seswarrajan, line 18: Kindly include **arm64** as well

src/packaging/discovery-engine.service
Ankurk99: Not sure if we need to run discovery by root. But from here https://stackoverflow.com/a/53537418, it seems the service is by default started in root mode. I think we can remove it then

src/packaging/post-install.sh
```diff
@@ -0,0 +1,4 @@
+set -e
+
+/bin/systemctl daemon-reload
+/bin/systemctl start knoxAutoPolicy.service
```
seswarrajan, line 4: Also if possible/required, add below line 

/bin/systemctl enable knoxAutoPolicy.service

seswarrajan (commented):
Kindly address review comments if possible or it can be handled in later release.

src/packaging/post-install.sh
Ankurk99: Enabling the service will automatically start the discovery-engine on each system startup, I think that might not be required by the user. Wdyt?

src/.goreleaser.yml
Ankurk99: I think we can add it once we support KubeArmor too. :grin:

src/packaging/post-install.sh
nyrahul: Yeah, lets not start by default.

nyrahul (approved):
LGTM

PR #529, opened 2022-08-22
change serviceaccount namespace to default

rksharma95:
we are no longer deploying the discovery-engine with the namespace as `explorer`.
this PR changes the service-account namespace to the `default`. 
Signed-off-by: rk <ramakant@accuknox.com>

### DIFF
--- a/deployments/helm/values-dev.yaml
+++ b/deployments/helm/values-dev.yaml
@@ -18,7 +18,7 @@ namespace: accuknox-dev-knoxautopolicy
 
 #serviceaccountnamespace
 namespace:
-  name: explorer
+  name: default
 
 volumeMounts:
   - mountPath: /conf
--- a/deployments/helm/values-prod.yaml
+++ b/deployments/helm/values-prod.yaml
@@ -18,7 +18,7 @@ namespace: accuknox-knoxautopolicy
 
 #serviceaccountnamespace
 namespace:
-  name: explorer
+  name: default
 
 volumeMounts:
   - mountPath: /conf
--- a/deployments/helm/values-snapshot.yaml
+++ b/deployments/helm/values-snapshot.yaml
@@ -18,7 +18,7 @@ namespace: accuknox-dev-knoxautopolicy
 
 #serviceaccountnamespace
 namespace:
-  name: explorer
+  name: default
 
 volumeMounts:
   - mountPath: /conf
--- a/deployments/k8s/deployment.yaml
+++ b/deployments/k8s/deployment.yaml
@@ -136,4 +136,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: knoxautopolicy
-  namespace: explorer
+  namespace: default


Ankurk99:
Can you please also make the suitable changes in the kubearmor-client error message to remove "explorer"? https://github.com/kubearmor/kubearmor-client/blob/main/discover/discover.go#L67 (it's also in summary and insight)

deployments/helm/values-dev.yaml
```diff
@@ -18,7 +18,7 @@ namespace: accuknox-dev-knoxautopolicy
 
 #serviceaccountnamespace
 namespace:
-  name: explorer
+  name: default
```
nyrahul, line 21:
```suggestion
  name: explorer
```

deployments/helm/values-prod.yaml
```diff
@@ -18,7 +18,7 @@ namespace: accuknox-knoxautopolicy
 
 #serviceaccountnamespace
 namespace:
-  name: explorer
+  name: default
```
nyrahul, line 21:
```suggestion
  name: explorer
```

deployments/helm/values-snapshot.yaml
```diff
@@ -18,7 +18,7 @@ namespace: accuknox-dev-knoxautopolicy
 
 #serviceaccountnamespace
 namespace:
-  name: explorer
+  name: default
```
nyrahul, line 21:
```suggestion
  name: explorer
```

deployments/k8s/deployment.yaml
```diff
@@ -136,4 +136,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: knoxautopolicy
-  namespace: explorer
+  namespace: default
```
nyrahul, line 139:
```suggestion
```

nyrahul (changes requested):

deployments/k8s/deployment.yaml
rksharma95: removing namespace entry leading to the error when applied
`The ClusterRoleBinding "knoxautopolicy" is invalid: subjects[0].namespace: Required value`

Ankurk99: Can we keep it explorer then and specify the `-n explorer` in the installation command?

Ankurk99:
@rksharma95 Is there any movement here? Do we have a solution for removing the serviceaccount namespace and let user apply it?

nyrahul:
> @rksharma95 Is there any movement here? Do we have a solution for removing the serviceaccount namespace and let user apply it?

@rksharma95 , we need to conclude on this. Can you please raise this during our regular discussion meeting? Thanks

rksharma95:
> @rksharma95 Is there any movement here? Do we have a solution for removing the serviceaccount namespace and let user apply it?

can be handled in the case of Helm for k8s yaml deployment we have to use specific value of the namespace.

seswarrajan:
Closing this PR as the default ns for disocvery-engine is decided as accuknox-agents

PR #531, opened 2022-08-23
Add fromsource, ns filters in system policy

seswarrajan:

### DIFF
--- a/src/config/configManager.go
+++ b/src/config/configManager.go
@@ -143,17 +143,17 @@ func LoadConfigFromFile() {
 		NetSkipCertVerification: viper.GetBool("application.network.skip-cert-verification"),
 	}
 
-	var ns, notNs []string
-	namespaces := viper.GetStringSlice("application.network.namespace-filter")
-	for _, n := range namespaces {
+	var nw_ns, nw_notNs []string
+	nw_namespaces := viper.GetStringSlice("application.network.namespace-filter")
+	for _, n := range nw_namespaces {
 		if n[0] == '!' {
-			notNs = append(notNs, n[1:])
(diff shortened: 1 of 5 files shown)


src/config/configManager.go
```diff
@@ -176,6 +176,19 @@ func LoadConfigFromFile() {
...
 
+	var sys_ns, sys_notNs []string
+	sys_namespaces := viper.GetStringSlice("application.system.namespace-filter")
+	for _, n := range sys_namespaces {
+		if n[0] == '!' {
+			sys_notNs = append(sys_notNs, n[1:])
+		} else {
+			sys_ns = append(sys_ns, n)
+		}
+	}
```
nyrahul: Can we refactor the code ... between application and network? Seems we can create a function to read the filter and return the ns and notNs vars.

src/plugin/kubearmor.go
```diff
@@ -432,6 +456,26 @@ func StartKubeArmorRelay(StopChan chan struct{}, cfg types.ConfigKubeArmorRelay)
...
+						}
+					}
+				}
+
+				for _, srcFilter := range fromSourceFilter {
+					if strings.Contains(res.Source, srcFilter) {
+						continue
+					}
+				}
+
```
nyrahul: This code seems redundant to what is available before. Can you refactor and create a function to handle this?

nyrahul (changes requested):

wazir-ahmed (approved):
Other than @nyrahul's comment, I don't have anything new to add. The PR looks good to me.

nyrahul (approved):
LGTM 👍🏼

PR #536, opened 2022-08-26
Enable Observability 

seswarrajan:


src/conf/local-file.yaml
```diff
@@ -37,7 +37,7 @@ observability:
   cron-job-time-interval: "0h0m10s"         # format: XhYmZs
   dbname: ./accuknox-obs.db
   system-observability: true
-  network-observability: true
+  network-observability: false
```
nyrahul: why disable it?

Ankurk99: do we want to keep network observability off by default?

src/libs/mysqlHandler.go
```diff
@@ -908,24 +908,27 @@ func UpdateWorkloadProcessFileSetMySQL(cfg types.ConfigDB, wpfs types.WorkloadPr
...
 
 // UpdateOrInsertKubearmorLogsSQLite -- Update existing log or insert a new log into DB
-func UpdateOrInsertKubearmorLogsMySQL(cfg types.ConfigDB, kubearmorlogs []types.KubeArmorLog) error {
+func UpdateOrInsertKubearmorLogsMySQL(cfg types.ConfigDB, kubearmorlogmap map[types.KubeArmorLog]int) error {
 	db := connectMySQL(cfg)
 	defer db.Close()
 
-	for _, kubearmorlog := range kubearmorlogs {
-		if err := updateOrInsertKubearmorLogMySQL(db, kubearmorlog); err != nil {
+	start := time.Now().UnixNano() / int64(time.Millisecond)
```
Ankurk99: Instead of deriving, can we directly use time.Now().UnixMicro() for time in microseconds? or time.Now().Unix() for seconds.

src/libs/mysqlHandler.go
```diff
@@ -961,11 +961,10 @@ func updateOrInsertKubearmorLogMySQL(db *sql.DB, kubearmorlog types.KubeArmorLog
 
 	if err == nil && rowsAffected == 0 {
 
-		updateQueryString := `(cluster_name,host_name,namespace_name,pod_name,container_id,container_name,
-		uid,type,source,operation,resource,labels,data,category,action,start_time,
-		updated_time,result,total) values(?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)`
```
Ankurk99, line 966: why are we removing log type?

src/libs/mysqlHandler.go
```diff
@@ -1452,23 +1358,23 @@ func GetCiliumLogsMySQL(cfg types.ConfigDB, filterLog types.CiliumLog) ([]types.
...
+	for _, ciliumLog := range ciliumlogs {
+		if err := updateOrInsertCiliumLogMySQL(db, ciliumLog); err != nil {
 			log.Error().Msg(err.Error())
 		}
 	}
 	return err
 }
 
-// UpdateCiliumLogsMySQL -- Update existing log with time and count
 func updateOrInsertCiliumLogMySQL(db *sql.DB, ciliumlog types.CiliumLog) error {
```
Ankurk99, line 1375: these two function names are a bit confusing, can we update these?

src/observability/kubearmor.go
```diff
@@ -48,7 +78,7 @@ func ProcessSystemLogs() {
 	SystemLogsMutex.Unlock()
 
 	ObsMutex.Lock()
-	defer ObsMutex.Unlock()
+	//defer ObsMutex.Unlock()
```
Ankurk99: we can remove it.

src/protobuf/v1/observability/observability.proto
```diff
@@ -4,56 +4,59 @@ package v1.observability;
...
-    string addressFamily = 1;
-    string path = 2;
+message PodNameResponse {
+    repeated string PodName = 1;
 }
-
-message OutgoingServerConnections {
-    string addressFamily = 1;
-    string path = 2;
-}
```
Ankurk99, line 59: Can you please terminate the file with an empty line?

src/conf/local-file.yaml
seswarrajan: This is kept off temporarily until full validation. Once validation is complete, will be turned ON.

seswarrajan: Set it to TRUE now

src/libs/mysqlHandler.go
seswarrajan: I think we do not need Log Type
As we are handling/see difference in pod names 
vm/container have vm-podname and container-podname

src/libs/mysqlHandler.go
seswarrajan: Can you elaborate more. Why is this confusing ?

nyrahul (approved):

src/conf/local-file.yaml
Ankurk99: https://github.com/accuknox/discovery-engine/pull/536#discussion_r957022237

src/libs/mysqlHandler.go
Ankurk99: one function is `updateOrInsertCiliumLogMySQL` and then above at line 1362 we have `UpdateOrInsertCiliumLogsMySQL()` both have similar names.

src/conf/local-file.yaml
seswarrajan: Enabled it now @nyrahul

Ankurk99 (approved):

PR #538, opened 2022-08-30
Modified hubble filter flags to get the least amount of traffic possible

wazir-ahmed:
Signed-off-by: Wazir Ahmed <wazir@accuknox.com>


nyrahul (approved):
LGTM 👍🏼 
What I see is:
1. No UDP response events will be considered
2. Only TCP SYN events will be considered (not even SYN-ACK)

wazir-ahmed:
@nyrahul Yes, you're correct.

PR #539, opened 2022-09-01
Enable observability/store AD policy in DB

seswarrajan:
1. Enable Observability
2. Add AD policy yaml to DB(policy_table) (https://kloudone.atlassian.net/browse/AK-10946) 

### DIFF
--- a/src/libs/mysqlHandler.go
+++ b/src/libs/mysqlHandler.go
@@ -721,7 +721,7 @@ func CreatePolicyTableMySQL(cfg types.ConfigDB) error {
 			"	`namespace` varchar(50) DEFAULT NULL," +
 			"	`labels` text DEFAULT NULL," +
 			"	`policy_name` varchar(150) DEFAULT NULL," +
-			"	`policy_data` text DEFAULT NULL," +
+			"	`policy_yaml` text DEFAULT NULL," +
 			"	`updated_time` bigint NOT NULL," +
 			"	PRIMARY KEY (`id`)" +
 			"  );"
@@ -1642,7 +1642,7 @@ func updateOrInsertPolicyMySQL(policy types.Policy, db *sql.DB) error {
 	var err error
 	queryString := ` policy_name = ? `
 
-	query := "UPDATE " + PolicyTable_TableName + " SET policy_data=?, updated_time=? WHERE " + queryString + " "
+	query := "UPDATE " + PolicyTable_TableName + " SET policy_yaml=?, updated_time=? WHERE " + queryString + " "
 
 	updateStmt, err := db.Prepare(query)
 	if err != nil {
@@ -1651,7 +1651,7 @@ func updateOrInsertPolicyMySQL(policy types.Policy, db *sql.DB) error {
 	defer updateStmt.Close()
 
 	result, err := updateStmt.Exec(
-		policy.PolicyData,
+		policy.PolicyYaml,
 		ConvertStrToUnixTime("now"),
 		policy.PolicyName,
 	)
@@ -1665,7 +1665,7 @@ func updateOrInsertPolicyMySQL(policy types.Policy, db *sql.DB) error {
 	if err == nil && rowsAffected == 0 {
 
 		insertStmt, err := db.Prepare("INSERT INTO " + PolicyTable_TableName +
-			"(type,cluster_name,namespace,labels,policy_name,policy_data,updatedtime) values(?,?,?,?,?,?,?)")
+			"(type,cluster_name,namespace,labels,policy_name,policy_yaml,updatedtime) values(?,?,?,?,?,?,?)")
 		if err != nil {
 			return err
 		}
@@ -1677,7 +1677,7 @@ func updateOrInsertPolicyMySQL(policy types.Policy, db *sql.DB) error {
 			policy.Namespace,
 			policy.Labels,
 			policy.PolicyName,
-			policy.PolicyData,
+			policy.PolicyYaml,
 			ConvertStrToUnixTime("now"),
 		)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


src/libs/mysqlHandler.go
```diff
@@ -706,6 +707,29 @@ func CreateTableNetworkLogsMySQL(cfg types.ConfigDB) error {
...
+
+	query :=
+		"CREATE TABLE IF NOT EXISTS `" + tableName + "` (" +
+			"	`id` INTEGER AUTO_INCREMENT," +
+			"	`type` varchar(50) DEFAULT NULL," +
+			"	`cluster_name` varchar(50) DEFAULT NULL," +
+			"	`namespace` varchar(50) DEFAULT NULL," +
+			"	`labels` text DEFAULT NULL," +
+			"	`policy_name` varchar(150) DEFAULT NULL," +
+			"	`policy_data` text DEFAULT NULL," +
```
nyrahul: Lets use `policy_yaml` as discussed.

src/libs/mysqlHandler.go
```diff
@@ -1595,3 +1619,69 @@ func GetPodNamesMySQL(cfg types.ConfigDB, filter types.ObsPodDetail) ([]string,
...
+		}
+		defer insertStmt.Close()
+
+		_, err = insertStmt.Exec(
+			policy.Type,
+			policy.Clustername,
+			policy.Namespace,
+			policy.Labels,
+			policy.PolicyName,
+			policy.PolicyData,
```
nyrahul: Lets call this PolicyYAML

nyrahul (changes requested):
Please fix all the lint warnings. Thanks

nyrahul:
@wazir-ahmed , please provide your review for this. Thanks

nyrahul (approved):
LGTM

src/systempolicy/systemPolicy.go
```diff
@@ -985,6 +986,8 @@ func updateSysPolicies() {
 
 	wpfsPolicies := populateKnoxSysPolicyFromWPFSDb("", "", "", "")
 
+	go insertSysPoliciesYamlToDB(wpfsPolicies)
```
nyrahul: Shouldn't this API be called below after checking for duplicate policies?
Secondly, why the use of go routine?
In fact it would be better to call this API from inside `libs.InsertSystemPolicies()`.

src/systempolicy/systemPolicy.go
```diff
@@ -1324,6 +1326,49 @@ func GenFileSetForAllPodsInCluster(clusterName string, pods []types.Pod, settype
 	return status
 }
 
+func insertSysPoliciesYamlToDB(policies []types.KnoxSystemPolicy) {
+	kubeArmorPolicies := plugin.ConvertKnoxSystemPolicyToKubeArmorPolicy(policies)
+
+	if len(kubeArmorPolicies) <= 0 {
+		return
+	}
+
```
nyrahul:
```suggestion
```

nyrahul (changes requested):

src/systempolicy/systemPolicy.go
seswarrajan: @nyrahul It is not possible to call the insertSysPoliciesYamlToDB() inside libs.InsertSystemPolicies() as this will cause circular import error (importing libs in systempolicy and systempolicy in libs). 
Are you suggesting to implement the whole function under DBHandler ? 

This API is called inside updateSysPolicy() and I hope it to not have any duplicate policies.

src/networkpolicy/networkPolicy.go
```diff
@@ -2133,6 +2137,52 @@ func PopulateNetworkPoliciesFromNetworkLogs(networkLogs []types.KnoxNetworkLog)
...
+			continue
+		}
+		yamlBytes, err := yaml.JSONToYAML(jsonBytes)
+		if err != nil {
+			log.Error().Msg(err.Error())
+			continue
+		}
+		policyYaml := string(yamlBytes)
+
+		for k, v := range ciliumPolicy.Spec.EndpointSelector.MatchLabels {
```
wazir-ahmed: VM/host policies should take into account here, For host policies, the labels will be available in `ciliumPolicy.Spec.NodeSelector.MatchLabels`. You should check which `MatchLabels` is empty.

src/types/policyData.go
```diff
@@ -309,3 +309,15 @@ type KubeArmorPolicy struct {
 
 	Spec KnoxSystemSpec `json:"spec,omitempty" yaml:"spec,omitempty"`
 }
+
+// =============== //
+// == Policy DB == //
+// =============== //
+type Policy struct {
+	Type        string
```
wazir-ahmed: IMO, we should add another field - `Kind`. Even though the `Type` specifies whether it is `system` or `network`. Within `system` policies, we have `KubeArmorPolicy` and `KubeArmorHostPolicy`. Likewise for the network - `CNP` and `CCNP`. So having `Kind` here will help us properly parse the contents of `PolicyYaml`.

src/libs/mysqlHandler.go
```diff
@@ -18,6 +18,7 @@ const TableNetworkPolicy_TableName = "network_policy"
 const TableSystemPolicy_TableName = "system_policy"
 const TableSystemLogs_TableName = "system_logs"
 const TableNetworkLogs_TableName = "network_logs"
+const PolicyTable_TableName = "policy_table"
```
wazir-ahmed: A table with name `policy_"table"`? :slightly_smiling_face:

wazir-ahmed (changes requested):
Overall the changes looks very good. But I have some comments..

src/libs/mysqlHandler.go
seswarrajan: I know. I cannot come up with any other name. We can say maybe policy_yaml_table.

wazir-ahmed: I can understand the struggle :slightly_smiling_face: | `policy_yaml` should be fine. Let's not add `table` as the suffix.

wazir-ahmed (approved):
Looks good to me :+1:

nyrahul (approved):
LGTM 👍🏼

PR #541, opened 2022-09-02
Handle summary request type

seswarrajan:


Ankurk99:
Can you please squash the commits, and I think we are good to merge

src/observability/summary.go
```diff
@@ -11,7 +12,8 @@ func GetSummaryData(request *opb.Request) (*opb.Response, error) {
 	resp := opb.Response{}
 	var err error = nil
 
-	if request.Type == "system" || request.Type == "all" {
+	if strings.Contains(request.Type, "process") || strings.Contains(request.Type, "file") || strings.Contains(request.Type, "network") {
```
Ankurk99, line 15: can the request.Type be set to "all"? Then this check won't match, right?

src/types/observability.go
```diff
@@ -160,21 +160,14 @@ type SysObsNwData struct {
...
-	DestinationLabel     string
+	SrcPodName           string
+	DestPodName          string
 	DestinationNamespace string
+	DestinationLabel     string
 	Protocol             string
 	Port                 string
 	Status               string
-	Count                int
+	Count                string
```
Ankurk99, line 171: May I know the reason for converting Count to string?

src/observability/summary.go
seswarrajan: No @Ankurk99 
From karmor summary, I am planning to pass the request type as 
**"process,file,network"**

src/types/observability.go
seswarrajan: I am sending this data to protobuf as a string, hence the basic storage type us changed to string.

src/observability/summary.go
Ankurk99: As we discussed, the default type will be all, i.e "process,file,network" so the user won't need to pass "all".

Ankurk99 (approved):
LGTM :+1:

wazir-ahmed (approved):
LGTM.

PR #569, opened 2022-10-19
Initial commit for summarizer and publisher (worker DE)

seswarrajan:


src/observability/consumer.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+// RelaySummaryEventToGrpcStream
+func RelaySummaryEventToGrpcStream(stream grpc.ServerStream, consumer *SummaryConsumer) error {
+	for {
+		select {
+		case <-stream.Context().Done():
+			// client disconnected
+			RemoveConsumer(SummaryConsumer, consumer)
```
rajaSahil:
```suggestion
			SysSummaryStore.RemoveConsumer(consumer)
```

rajaSahil (changes requested):

rajaSahil (approved):

src/conf/local-file.yaml
```diff
@@ -2,7 +2,7 @@ application:
   name: knoxautopolicy
   network:
     operation-mode: 1                             # 1: cronjob | 2: one-time-job
-    operation-trigger: 5
+    operation-trigger: 1
```
nyrahul: Operation trigger of 1 or 5 is too low

nyrahul (approved):

src/conf/local-file.yaml
seswarrajan: @nyrahul In case of neteork opertaion for discoevry-engine, we are not processing all logs. We are processing only bare minimum logs which contains TCP and UDP related data. In that case, we need a minimum trigger of 5 to start the processing.

PR #573, opened 2022-10-26
Publisher: Fix publish logic in code

seswarrajan:


src/observability/consumer.go
```diff
@@ -19,11 +19,20 @@ type SummaryConsumer struct {
 	Events        chan *types.SystemSummary
 }
 
-type SummaryStore struct {
+type Store struct {
```
rajaSahil: nit: Can we define all global variables for observability in a struct and initialize it?

rajaSahil: nit: Can we rename `Store` to `ConsumersStore`?

src/observability/consumer.go
```diff
@@ -19,11 +19,20 @@ type SummaryConsumer struct {
...
 	Consumers map[*SummaryConsumer]struct{}
 	Mutex     sync.Mutex
 }
 
+var SummaryStore Store
```
rajaSahil: nit: Can we rename `SummaryStore` to `SysSummary`?

src/observability/consumer.go
```diff
@@ -38,22 +47,24 @@ func NewSummaryConsumer(req *ppb.SummaryRequest) *SummaryConsumer {
...
+func (sc *Store) RemoveConsumer(c *SummaryConsumer) {
 	sc.Mutex.Lock()
 	defer sc.Mutex.Unlock()
 
+	log.Info().Msgf("Consumer removed")
 	delete(sc.Consumers, c)
 }
 
-func (sc *SummaryStore) Publish(summary *types.SystemSummary) {
+func (sc *Store) Publish(summary *types.SystemSummary) {
```
rajaSahil: Please add comments.

src/observability/consumer.go
seswarrajan: We can use as SysSummary

src/observability/consumer.go
seswarrajan: That will look good @rajaSahil 
We got some code clean up to do. We can take this up as part of that task.

rajaSahil (approved):
LGTM, Thanks!

PR #575, opened 2022-10-28
Add publisher config to deployment yaml

seswarrajan:

### DIFF
--- a/src/systempolicy/systemPolicy.go
+++ b/src/systempolicy/systemPolicy.go
@@ -168,7 +168,6 @@ type SysLogKey struct {
 // ================ //
 
 func getSystemLogs() []types.KnoxSystemLog {
-	cfgDB := types.ConfigDB{}
 	systemLogs := []types.KnoxSystemLog{}
 
 	if SystemLogFrom == "file" {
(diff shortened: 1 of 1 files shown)


deployments/k8s/deployment.yaml
```diff
@@ -69,7 +69,12 @@ data:
       cron-job-time-interval: "0h0m10s"         # format: XhYmZs
       dbname: ./accuknox-obs.db
       system-observability: true
-      network-observability: true
+      network-observability: false
```
Ankurk99: why are we disabling network-observability?

seswarrajan: @Ankurk99 We do not have network observability. We will be using kubearmor to get network observability and generic network logs + observability as well.

rajaSahil: @seswarrajan  `network-observability` is not being used anywhere in the code?

Ankurk99: Ah! so have we removed the cilium network observability support?

seswarrajan: We are storing logs into network_logs DB and also retreiving it for observability. But not being used. 
Which will be removed in future.

Ankurk99 (approved):
LGTM

PR #580, opened 2022-11-02
K8s Net pol -- UDP handling

seswarrajan:

### DIFF
--- a/src/networkpolicy/networkPolicy.go
+++ b/src/networkpolicy/networkPolicy.go
@@ -2222,6 +2222,7 @@ func writeNetworkPoliciesYamlToDB(policies []types.KnoxNetworkPolicy) {
 		k8sNetPolicies := plugin.ConvertKnoxNetPolicyToK8sNetworkPolicy("", "", policies)
 
 		for i, np := range k8sNetPolicies {
+			np.ClusterName = ""
 			jsonBytes, err := json.Marshal(np)
 			if err != nil {
 				log.Error().Msg(err.Error())
--- a/src/networkpolicy/networkPolicy.go
+++ b/src/networkpolicy/networkPolicy.go
@@ -1761,6 +1761,54 @@ func mergeHttpRules(existRule types.L47Rule, newRule types.L47Rule) (bool, bool,
 	return false, false, nil
 }
 
+func popultaeIngressEgressPolicyFromKnoxNetLog(log *types.KnoxNetworkLog, pods []types.Pod) types.KnoxNetworkPolicy {
+	iePolicy := types.KnoxNetworkPolicy{}
+
+	if log.Direction == "EGRESS" {
+		iePolicy = buildNewKnoxEgressPolicy()
+		egress := types.Egress{}
+
+		iePolicy.Spec.Selector.MatchLabels = getEndpointMatchLabels(log.SrcPodName, pods)
+
+		var cidrs []string
+		cidrs = append(cidrs, "0.0.0.0/32")
+
+		egress.ToCIDRs = append(egress.ToCIDRs, types.SpecCIDR{
+			CIDRs: cidrs,
+		})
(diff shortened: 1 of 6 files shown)


src/networkpolicy/networkPolicy.go
```diff
@@ -1598,13 +1598,31 @@ func mergeIngressPolicies(existPolicy types.KnoxNetworkPolicy, policies []types.
 						}
 					}
 				}
+			} else if len(newIngress.FromCIDRs) > 0 && len(newIngress.ToPorts) > 0 {
```
nyrahul, line 1633: In this condition only `FromCIDRs` at index 0 is checked. What happens if we get multiple `FromCIDRs` in the rule.

src/networkpolicy/networkPolicy.go
```diff
@@ -1668,7 +1686,24 @@ func mergeEgressPolicies(existPolicy types.KnoxNetworkPolicy, policies []types.K
 						}
 					}
 				}
+			} else if len(newEgress.ToCIDRs) > 0 && len(newEgress.ToPorts) > 0 {
```
nyrahul, line 1708: This is duplicate code. Please keep an API for CIDR matching and call from both the ToCIDRs and FromCIDRs condition.

src/networkpolicy/networkPolicy.go
```diff
@@ -1858,6 +1893,49 @@ func convertKnoxNetworkLogToKnoxNetworkPolicy(log *types.KnoxNetworkLog, pods []
...
+
+		egress.ToPorts = append(egress.ToPorts, types.SpecPort{
+			Port:     strconv.Itoa(log.DstPort),
+			Protocol: libs.GetProtocol(log.Protocol),
+		})
+
+		ePolicy.Spec.Egress = append(ePolicy.Spec.Egress, egress)
+		ePolicy.Metadata["namespace"] = log.SrcNamespace
+		egressPolicy = &ePolicy
+	} else if log.DstPodName == "" && len(log.DstReservedLabels) == 0 && log.Direction == "INGRESS" {
```
nyrahul: There is too much of duplicate code in the two condition. Can we create a function to handle this?

src/plugin/k8sNetwork.go
```diff
@@ -31,32 +28,52 @@ func ConvertKnoxNetPolicyToK8sNetworkPolicy(clustername, namespace string) []nv1
...
+				} else {
+					port = nv1.NetworkPolicyPort{
+						Protocol: &protocol,
+					}
+				}
+
+				if len(eg.ToCIDRs) > 0 {
+					to = nv1.NetworkPolicyPeer{
+						IPBlock: &nv1.IPBlock{
+							CIDR: eg.ToCIDRs[0].CIDRs[0],
```
nyrahul: What happens if `len(eg.ToCIDRs) > 1`?

src/plugin/k8sNetwork.go
```diff
@@ -71,27 +88,46 @@ func ConvertKnoxNetPolicyToK8sNetworkPolicy(clustername, namespace string) []nv1
...
+						},
+						Protocol: &protocol,
+					}
+				} else {
+					port = nv1.NetworkPolicyPort{
+						Protocol: &protocol,
+					}
+				}
+
+				if len(ing.FromCIDRs) > 0 {
```
nyrahul: Again there is duplication of code here from ToCIDR and FromCIDR perspective.

nyrahul (changes requested):

src/networkpolicy/networkPolicy.go
seswarrajan: Each egress rule contains only one CIDRs. Hence no issue.

nyrahul (approved):

PR #583, opened 2022-11-07
make network rules optional for system policy

achrefbensaad:
This PR disable the generation of network rules for system policies via grpc.
the user can optionally generate them by setting **includenetwork** to true.


seswarrajan (approved):
LGTM

PR #595, opened 2022-11-18
Purge Old DB Entries

yasin-cs-ko-ak:
- This PR contains generic config for purge old db entries for specific time. 
- It uses cron job to run the db query.
- discovery-engine currently uses MySQL and SQLite as databases.
- `accuknox-obs.db` and `accuknox-pol.db` added in `local-file.yaml` as of now.

### DIFF
--- a/src/libs/dbHandler.go
+++ b/src/libs/dbHandler.go
@@ -3,8 +3,12 @@ package libs
 import (
 	"database/sql"
 	"errors"
+	"sync"
 
+	cfg "github.com/accuknox/auto-policy-discovery/src/config"
+	logger "github.com/accuknox/auto-policy-discovery/src/logging"
 	"github.com/accuknox/auto-policy-discovery/src/types"
+	"github.com/robfig/cron"
 )
 
 // ================= //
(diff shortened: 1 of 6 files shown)


nyrahul (approved):
LGTM 👍

seswarrajan:
Thanks @nyrahul for the approval. I will test once and will merge this PR.

src/libs/dbHandler.go
```diff
@@ -616,3 +620,49 @@ func getSysSummarySQL(db *sql.DB, dbName string, filterOptions types.SystemSumma
...
+var (
+	CfgDB                                      types.ConfigDB
+	SystemLogsMutex, NetworkLogsMutex, DBMutex *sync.Mutex
+	PurgeDBCronJob                             *cron.Cron
+	PurgeDBMap                                 types.ConfigPurgeOldDBEntries
+)
+
+func initMutex() {
+	DBMutex = &sync.Mutex{}
+	if cfg.GetCfgPurgeOldDBEntriesEnable() {
```
rajaSahil: Why are we using two if statements for same condition?

src/libs/dbHandler.go
```diff
@@ -616,3 +620,49 @@ func getSysSummarySQL(db *sql.DB, dbName string, filterOptions types.SystemSumma
...
+// ==================================== //
+var (
+	CfgDB                                      types.ConfigDB
+	SystemLogsMutex, NetworkLogsMutex, DBMutex *sync.Mutex
+	PurgeDBCronJob                             *cron.Cron
+	PurgeDBMap                                 types.ConfigPurgeOldDBEntries
+)
+
+func initMutex() {
+	DBMutex = &sync.Mutex{}
```
seswarrajan: Why are we defining 2 mutexes. I did not see them being used anywhere else.

src/libs/dbHandler.go
```diff
@@ -616,3 +620,49 @@ func getSysSummarySQL(db *sql.DB, dbName string, filterOptions types.SystemSumma
...
+			return
+		}
+		PurgeDBCronJob.Start()
+		log.Info().Msg("Purging Old DB Entries cron job started")
+	}
+}
+
+func PurgeOldDBEntriesCronJob() {
+	if cfg.GetCfgPurgeOldDBEntriesEnable() {
+		go PurgeOldDBEntriesMySQL(types.ConfigDB{})
```
rajaSahil: How are we ensuring, these go routines executes? Shouldn't be there wait() for these goroutines?

src/libs/dbHandler.go
seswarrajan: We are defining DB mutex and not using it as lock for deleting the DB entries.

seswarrajan (changes requested):

src/libs/mysqlHandler.go
```diff
@@ -1724,3 +1725,26 @@ func GetSystemSummaryMySQL(cfg types.ConfigDB, filterOptions types.SystemSummary
 
 	return res, err
 }
+
+// ========================== //
+// == Purge Old DB Entries == //
+// ========================== //
+
+func PurgeOldDBEntriesMySQL(cfg types.ConfigDB) error {
```
rajaSahil: We are returning error, but where it is being handled? 
And since this function is called through a seperate goroutine, should we return error?

src/libs/sqliteHandler.go
```diff
@@ -1728,3 +1728,25 @@ func GetSystemSummarySQLite(cfg types.ConfigDB, filterOptions types.SystemSummar
 
 	return res, err
 }
+
+// ========================== //
+// == Purge Old DB Entries == //
+// ========================== //
+func PurgeOldDBEntriesSQLite(cfg types.ConfigDB) error {
```
rajaSahil: Same comment as above https://github.com/kubearmor/discovery-engine/pull/595#discussion_r1027590493

PR #596, opened 2022-11-21
Read cluster_id from env variable and add it to summary data

seswarrajan:

### DIFF
--- a/src/observability/consumer.go
+++ b/src/observability/consumer.go
@@ -1,7 +1,6 @@
 package observability
 
 import (
-	"strconv"
 	"sync"
 
 	ppb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/publisher"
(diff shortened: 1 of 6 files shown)


src/observability/consumer.go
```diff
@@ -109,10 +108,9 @@ func validateSummaryRequest(consumer *SummaryConsumer, summary types.SystemSumma
 
 func convertSystemSummaryToGrpcResponse(summary *types.SystemSummary) *ppb.SummaryResponse {
 
-	clusterId, _ := strconv.ParseInt(summary.ClusterId, 0, 32)
 	return &ppb.SummaryResponse{
 		ClusterName:    summary.ClusterName,
-		ClusterId:      int32(clusterId),
+		ClusterId:      summary.ClusterId,
```
rajaSahil, line 113: I can see some inconsistency with data types for ID's i.e(` cluster id: string, workspace id: int`). Any reason? Can we follow the same data types for all ID's?

seswarrajan: I think it is not inconsistent. The previous/old changes are inconsistent. Now we have cluster & workspace ID as int.

rajaSahil (approved):
LGTM, Thanks!

PR #597, opened 2022-11-21
Purge Old DB Entries

yasin-cs-ko-ak:
• This PR contains generic config for purge old db entries for specific time.
• It uses cron job to run the db query.
• discovery-engine currently uses MySQL and SQLite as databases.
• `accuknox-obs.db` and `accuknox-pol.db` added in `local-file.yaml` as of now.

### DIFF
--- a/src/libs/dbHandler.go
+++ b/src/libs/dbHandler.go
@@ -4,7 +4,10 @@ import (
 	"database/sql"
 	"errors"
 
+	cfg "github.com/accuknox/auto-policy-discovery/src/config"
+	logger "github.com/accuknox/auto-policy-discovery/src/logging"
 	"github.com/accuknox/auto-policy-discovery/src/types"
+	"github.com/robfig/cron"
 )
 
 // ================= //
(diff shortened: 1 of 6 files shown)


src/libs/mysqlHandler.go
```diff
@@ -1724,3 +1725,24 @@ func GetSystemSummaryMySQL(cfg types.ConfigDB, filterOptions types.SystemSummary
...
+	defer db.Close()
+
+	timeNow := (ConvertStrToUnixTime("now"))
+	purgeTime := (config.GetCfgPublisherCronJobTime()) //sec
+	PurgeTimeValue, err := strconv.ParseInt(purgeTime, 10, 64)
+	if err != nil {
+		log.Error().Msg(err.Error())
+	}
+	ConvertedValue := timeNow - PurgeTimeValue
+	query := "DELETE FROM system_summary WHERE updated_time < " + strconv.Itoa(int(ConvertedValue))
```
rajaSahil, line 1744: Shouldn't we add mutex lock while deleting the table?

src/libs/sqliteHandler.go
```diff
@@ -1728,3 +1728,23 @@ func GetSystemSummarySQLite(cfg types.ConfigDB, filterOptions types.SystemSummar
...
+	defer db.Close()
+
+	timeNow := (ConvertStrToUnixTime("now"))
+	purgeTime := (config.GetCfgPublisherCronJobTime()) //sec
+	PurgeTimeValue, err := strconv.ParseInt(purgeTime, 10, 64)
+	if err != nil {
+		log.Error().Msg(err.Error())
+	}
+	ConvertedValue := timeNow - PurgeTimeValue
+	query := "DELETE FROM system_summary WHERE updated_time < " + strconv.Itoa(int(ConvertedValue))
```
rajaSahil, line 1746: Same comment as above

src/libs/mysqlHandler.go
yasin-cs-ko-ak: It is not necessary here. Because It's happening in specific time.

src/libs/sqliteHandler.go
yasin-cs-ko-ak: It is not necessary here. Because It's happening in specific time.

src/libs/mysqlHandler.go
rajaSahil: @yasin-cs-ko-ak Are we doing any other db operation for same table in any of the goroutines? 
cc: @seswarrajan

yasin-cs-ko-ak: @rajaSahil There is no need of goroutines. So I removed it. 
cc: @seswarrajan

src/libs/dbHandler.go
```diff
@@ -616,3 +619,36 @@ func getSysSummarySQL(db *sql.DB, dbName string, filterOptions types.SystemSumma
...
+			return
+		}
+		PurgeDBCronJob.Start()
+		log.Info().Msg("Purging Old DB Entries cron job started")
+	}
+}
+
+func PurgeOldDBEntriesCronJob() {
+	if cfg.GetCfgPurgeOldDBEntriesEnable() {
+		PurgeOldDBEntriesMySQL(types.ConfigDB{})
```
seswarrajan, line 652: Check the config choice of DB to use from conf/local-file.yaml and then call in a if else statement

seswarrajan: if MySQL{
PurgeOldDBEntriesMySQL()
} else if SQLite {
PurgeOldDBEntriesSQLite()
}

seswarrajan (changes requested):

src/libs/dbHandler.go
yasin-cs-ko-ak: @seswarrajan Changes are done.

src/libs/mysqlHandler.go
seswarrajan: Yes @rajaSahil We will be doing add/update/read operations on this DB in another cron

seswarrajan (commented):
Kindly respond to all comments by @rajaSahil

src/libs/mysqlHandler.go
yasin-cs-ko-ak: As of now we are only purging the data. There is no other operations on this db other than that. @rajasahil

yasin-cs-ko-ak: As of now we are only purging the data. There is no other operations on this db other than that. @rajaSahil

yasin-cs-ko-ak:
> Kindly respond to all comments by @rajaSahil

@seswarrajan All the comments are addressed. Thank you

PR #600, opened 2022-11-22
optimize auto-detected paths

achrefbensaad:
optimize auto-detected paths

### DIFF
--- a/src/common/pathAggregator.go
+++ b/src/common/pathAggregator.go
@@ -167,12 +167,17 @@ func buildPathTree(treeMap map[string]*Node, paths []string) {
 
 		rootPath := tokenizedPaths[0]
 
+		if rootPath == "/tmp" {
+			tokenizedPaths = pattern.FindAllString("/tmp/", -1)
+		}
+
 		if rootNode, ok := treeMap[rootPath]; !ok {
 			newRoot := &Node{
 				depth:      0,
 				path:       rootPath,
 				touchCount: 1,
 				childNodes: []*Node{},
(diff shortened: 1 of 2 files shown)


src/common/pathAggregator.go
```diff
@@ -167,12 +167,17 @@ func buildPathTree(treeMap map[string]*Node, paths []string) {
 
 		rootPath := tokenizedPaths[0]
 
+		if rootPath == "/tmp" {
```
seswarrajan, line 170: Why are we specifically adding tmp for check here @achrefbensaad ?

achrefbensaad: tmp files creates by applications under /tmp have random names and are not consistant between pod restarts, so we need to allow /tmp/**.
Adding tmp there is the simplest way to guarenty that behavior.

seswarrajan (approved):
LGTM

Ankurk99:
> tmp files creates by applications under /tmp have random names and are not consistant between pod restarts, so we need to allow /tmp/**.
Adding tmp there is the simplest way to guarenty that behaviour

But we have also seen this behaviour with the "var" folder too (both /var/ and /var) enteries exist, right?

achrefbensaad:
> But we have also seen this behaviour with the "var" folder too (both /var/ and /var) enteries exist, right?

Thats in the summary not in the wpfs

Ankurk99 (approved):
You may want to take a look at the go-lint warning. Otherwise LGTM :+1:

nyrahul (approved):

PR #602, opened 2022-11-23
Kind handling in grpc for network policy

seswarrajan:


rajaSahil (approved):

achrefbensaad (approved):
LGTM!

src/networkpolicy/helperFunctions.go
```diff
@@ -796,8 +810,9 @@ func GetNetPolicy(cluster, namespace, policyType string) *wpb.WorkerResponse {
 
 			response.Ciliumpolicy = append(response.Ciliumpolicy, &ciliumpolicy)
 		}
-		response.K8SNetworkpolicy = nil
-	} else if strings.Contains(policyType, "NetworkPolicy") {
+
+	}
+	if strings.Contains(pType, "generic") {
```
Ankurk99: `else if strings.Contains(pType, "generic") {`

seswarrajan: @Ankurk99 We are planning to get both at same point if option permits. That is why we are using a **if** case instead of **else if**

seswarrajan: @Ankurk99 We are planning to get both at same point if option permits. That is why we are using a **if** case instead of **else if**

src/networkpolicy/helperFunctions.go
```diff
@@ -778,8 +778,22 @@ func WriteNetworkPoliciesToFile(cluster, namespace string) {
 func GetNetPolicy(cluster, namespace, policyType string) *wpb.WorkerResponse {
 
 	var response wpb.WorkerResponse
+	var pType string = ""
```
nyrahul:
```suggestion
```

src/networkpolicy/helperFunctions.go
Ankurk99: Ah! I see. Then shouldn't we also keep `if and else` here too?
```
		if v == "CiliumNetworkPolicy" {
			pType += "cilium"
		} else if v == "NetworkPolicy" {
			pType += "generic"
		}
```

src/networkpolicy/helperFunctions.go
```diff
@@ -778,8 +778,22 @@ func WriteNetworkPoliciesToFile(cluster, namespace string) {
...
+
+	for _, v := range strings.Split(policyType, ",") {
+		if v == "CiliumNetworkPolicy" {
+			pType += "cilium"
+		} else if v == "NetworkPolicy" {
+			pType += "generic"
+		}
+	}
+
+	if strings.Contains(pType, "cilium") {
```
nyrahul:
```suggestion
	// Setting all policy data to nil
	response.K8SNetworkpolicy = nil
	response.Ciliumpolicy = nil
	response.Kubearmorpolicy = nil

    pt := strings.Split(policyType, ",")
	if slices.IndexFunc(pt, func(c string) bool { return c == "CiliumNetworkPolicy" }) > -1 {
```

src/networkpolicy/helperFunctions.go
nyrahul:
```suggestion
	if slices.IndexFunc(pt, func(c string) bool { return c == "NetworkPolicy" }) > -1 {
```

src/networkpolicy/helperFunctions.go
nyrahul: remember to import `"golang.org/x/exp/slices"`

nyrahul (changes requested):

src/networkpolicy/helperFunctions.go
seswarrajan: No @Ankurk99 We support to display both at same time

nyrahul (approved):

PR #603, opened 2022-11-23
Purge Old DB Entries

yasin-cs-ko-ak:
• This PR contains generic config for purge old db entries for specific time.
• It uses cron job to run the db query.
• discovery-engine currently uses MySQL and SQLite as databases.
• accuknox-obs.db and accuknox-pol.db added in local-file.yaml as of now.

### DIFF
--- a/src/libs/dbHandler.go
+++ b/src/libs/dbHandler.go
@@ -4,7 +4,10 @@ import (
 	"database/sql"
 	"errors"
 
+	cfg "github.com/accuknox/auto-policy-discovery/src/config"
+	logger "github.com/accuknox/auto-policy-discovery/src/logging"
 	"github.com/accuknox/auto-policy-discovery/src/types"
+	"github.com/robfig/cron"
 )
 
 // ================= //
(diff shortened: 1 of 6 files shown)


src/libs/dbHandler.go
```diff
@@ -616,3 +619,45 @@ func getSysSummarySQL(db *sql.DB, dbName string, filterOptions types.SystemSumma
...
+			log.Error().Msg(err.Error())
+			return
+		}
+		PurgeDBCronJob.Start()
+		log.Info().Msg("Purging Old DB Entries cron job started")
+	}
+}
+
+// Checking which type of Database
+func PurgeOldDBEntriesCronJob(cfg types.ConfigDB) {
```
seswarrajan: This is a cron job CB function and does not have an input. 
You need to read the config from config directly and process the same.

seswarrajan:
```suggestion
func PurgeOldDBEntriesCronJob() {
	if config.CurrentCfg.ConfigDB.ConfigDB == "mysql" {
		PurgeOldDBEntriesMySQL(types.ConfigDB{})
	} else if config.CurrentCfg.ConfigDB.ConfigDB == "sqlite3" {
		PurgeOldDBEntriesSQLite(types.ConfigDB{})
	}
```

seswarrajan (changes requested):

src/libs/dbHandler.go
yasin-cs-ko-ak: Resolved.

yasin-cs-ko-ak: I committed the changes.

seswarrajan (approved):
LGTM

PR #609, opened 2022-11-24
Updated Helm Chart

wazir-ahmed:
- Moved helm chart directory to `deployments/helm`.
- Updated pod resource request and limit.
- Updated `ClusterRole` with required `resources`.

Signed-off-by: Wazir Ahmed <wazir@accuknox.com>


seswarrajan (approved):
LGTM

deployments/helm/values.yaml
```diff
@@ -72,11 +72,11 @@ volumeMountsDiscoveryEngine:
...
 resourcesDiscoveryEngine:
   requests:
-    cpu: 800m
-    memory: 2Gi
+    cpu: 50m
+    memory: 20Mi
   limits:
-    cpu: 800m
-    memory: 2Gi
+    cpu: 500m
```
rajaSahil, line 78: Have we verified the limits? 
@wazir-ahmed Yesterday, we saw discovery engine using cpu and memory more than the values used here.

wazir-ahmed: Yes. But now in the same `accuknox-e2e-01` cluster, DE is running for more than two hours and only using a memory of 20MB and cpu of 300-400m core. That's why I configured the values slightly above currently utilized cpu/memory.

rajaSahil: Okay, but we might have to revisit and check it through stress testing.

rajaSahil (approved):

PR #610, opened 2022-11-24
make disocvered policies work

achrefbensaad:
Signed-off-by: Achref ben saad <achref@accuknox.com>

### DIFF
--- a/src/systempolicy/systemPolicy.go
+++ b/src/systempolicy/systemPolicy.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/clarketm/json"
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/yaml"
 
 	"github.com/accuknox/auto-policy-discovery/src/cluster"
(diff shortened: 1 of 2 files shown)


src/systempolicy/systemPolicy.go
```diff
@@ -788,6 +789,42 @@ func mergeSysPolicies(pols []types.KnoxSystemPolicy) []types.KnoxSystemPolicy {
...
+			if len(processDirRule.FromSource) == 0 {
+				gloablBinaries = append(gloablBinaries, processDirRule.Dir)
+			}
+		}
+	}
+
+	for _, pol := range results {
+		for i, filePath := range pol.Spec.File.MatchPaths {
+			for _, binary := range filePath.FromSource {
+				if slices.Contains(gloablBinaries, binary.Path) {
```
seswarrajan: It should be something like strings.HasPrefix() for each path right. Are we not comparing to full fromSource path ?

src/systempolicy/systemPolicy.go
```diff
@@ -788,6 +789,42 @@ func mergeSysPolicies(pols []types.KnoxSystemPolicy) []types.KnoxSystemPolicy {
 			})
 		}
 	}
+	// If a binary is allowed without fromSource, Convert the file access for it to global access
+
+	// Pre-compute: needed to optimize search
+
+	gloablBinaries := []string{}
```
Ankurk99: globalBinaries*

src/systempolicy/systemPolicy.go
achrefbensaad: We dont need that because fromsource has absolute binary paths.

seswarrajan (approved):
LGTM

achrefbensaad:
@seswarrajan , @Ankurk99  can you please re-review this.

PR #625, opened 2022-12-06
reduce the usage of k8s client api

Ankurk99:
- add pprof support 
- avoid the k8s client usage to once per function call instead of on per log basis
- Add optimization flags to go build (-O flag enables compiler's optimization, `-gcflags="-m"` enables the inlining of functions which can improve the performance of generated code) 
- refractor some code


src/plugin/kubearmor.go
```diff
@@ -554,6 +554,7 @@ func ConvertKubeArmorNetLogToKnoxNetLog(kaNwLogs []*pb.Log) []types.KnoxNetworkL
 		return nil
 	}
 
+	extractpodsvcflag := false
```
nyrahul: print the length(kaNwLogs)

nyrahul (changes requested):

src/libs/common.go
```diff
@@ -163,13 +166,42 @@ func (i *cfgArray) Set(str string) error {
...
 	flag.Var(&cmdlineCfg, "cfg", "Configuration key=val")
 
 	configFilePath := flag.String("config-path", "conf/", "conf/")
 	flag.Parse()
 
+	// Reset default routes (removing access to profiling)
+	http.DefaultServeMux = http.NewServeMux()
+
+	// Recreating routes to profiling manually
+	pprofServeMux := http.NewServeMux()
```
nyrahul: Can we add all the pprof http handling as part of pprofflag ?

nyrahul (changes requested):

src/libs/common.go
Ankurk99: Is it better now?

Ankurk99: We can enable for all default routes using DefaultServeMux() but that will expose the profiling tool on default server under /debug/pprof (thus the go-sec warning). Thus we are manually setting up the profiling routes. [docs.boostsecurity.io/rules/code-debugging-interface-publicly-exposed.html](https://docs.boostsecurity.io/rules/code-debugging-interface-publicly-exposed.html)

Ankurk99:
The most recent change increased the timeout limit to 90 seconds, this will allow us to do profiling of upto 90 seconds.

src/plugin/kubearmor.go
```diff
@@ -590,13 +591,19 @@ func ConvertKubeArmorNetLogToKnoxNetLog(kaNwLogs []*pb.Log) []types.KnoxNetworkL
 				continue
 			}
 
-			destPod, destLabels, destNs := cluster.ExtractPodSvcInfoFromIP(ip, kalog.ClusterName)
+			var destPod, destLabels, destNs string
```
nyrahul: Cannot have these vars inside of the for loop.

nyrahul (changes requested):

src/libs/common.go
```diff
@@ -159,17 +162,55 @@ func (i *cfgArray) Set(str string) error {
 	return nil
 }
 
+// Manually recreate routes for profiling
+func pprofinit() {
```
wazir-ahmed:
```suggestion
func pprofInit() {
```

src/libs/common.go
```diff
@@ -159,17 +162,55 @@ func (i *cfgArray) Set(str string) error {
...
+			log.Error().Msg("ListenAndServe: " + err.Error())
+		}
+	}()
+}
+
 /* configuration file values are final values */
 func CheckCommandLineConfig() {
 	var cmdlineCfg cfgArray
 
+	pprofflag := flag.Bool("pprof", false, "enable pprof")
```
wazir-ahmed:
```suggestion
	pprofFlag := flag.Bool("pprof", false, "enable pprof")
```

src/observability/summarizer.go
```diff
@@ -29,7 +29,14 @@ func extractNetworkInfoFromSystemLog(netLog pb.Log) (string, string, string, str
 		resslice := strings.Split(netLog.Resource, " ")
 		for _, locres := range resslice {
 			if strings.Contains(locres, "remoteip") {
-				ip, destLabel, destNs = cluster.ExtractPodSvcInfoFromIP(strings.Split(locres, "=")[1], netLog.ClusterName)
+				_, services, _, pods, err := cluster.GetAllClusterResources(netLog.ClusterName)
```
wazir-ahmed: `extractNetworkInfoFromSystemLog()` process an individual log. We shouldn't call `GetAllClusterResources()` here. That will give us the same performance as before. We should call `GetAllClusterResources()` function in any one of parent/grand-parent function, so that resources are fetched from k8s only once for processing all the buffered logs.

src/plugin/kubearmor.go
```diff
@@ -590,13 +596,23 @@ func ConvertKubeArmorNetLogToKnoxNetLog(kaNwLogs []*pb.Log) []types.KnoxNetworkL
 				continue
 			}
 
-			destPod, destLabels, destNs := cluster.ExtractPodSvcInfoFromIP(ip, kalog.ClusterName)
+			if existingClustername != kalog.ClusterName {
+				_, services, _, pods, err = cluster.GetAllClusterResources(kalog.ClusterName)
```
wazir-ahmed, line 599: As discussed, we should call `GetAllClusterResources()` in `DiscoverNetworkPolicyMain()` and use the svc/pod details we got from that function call in rest of the child functions. Calling `GetAllClusterResources()` here will not give us any performance improvement.

wazir-ahmed (changes requested):

src/plugin/kubearmor.go
Ankurk99: Actually we don't want to change the overall functionality as it may have some unforseen outcomes. The performance issue found in the pprof was particular to this function, that's why I only changed this function for now. 
Now regarding the performance improvement, here we are calling `GetAllClusterResources()` only once when we receive a unique clusterName from the logs. It's not called on each function, that's why am using the `existingClustername` variable.

wazir-ahmed: Okay got it. It's fine.

src/plugin/kubearmor.go
```diff
@@ -590,13 +596,23 @@ func ConvertKubeArmorNetLogToKnoxNetLog(kaNwLogs []*pb.Log) []types.KnoxNetworkL
...
+				if err != nil {
+					destPod = ip
+					destLabels = ""
+					destNs = ""
+				}
+				existingClustername = kalog.ClusterName
```
nyrahul:
```suggestion
				if err == nil {
     				existingClustername = kalog.ClusterName
				}
```

src/plugin/kubearmor.go
```diff
@@ -590,13 +596,23 @@ func ConvertKubeArmorNetLogToKnoxNetLog(kaNwLogs []*pb.Log) []types.KnoxNetworkL
...
+			}
 
+			destPod, destLabels, destNs = cluster.ExtractPodSvcInfoFromIP(ip, kalog.ClusterName, pods, services)
```
nyrahul:
```suggestion
			destPod, destLabels, destNs := cluster.ExtractPodSvcInfoFromIP(ip, kalog.ClusterName, pods, services)
```

src/plugin/kubearmor.go
```diff
@@ -556,6 +556,12 @@ func ConvertKubeArmorNetLogToKnoxNetLog(kaNwLogs []*pb.Log) []types.KnoxNetworkL
 
 	results := []types.KnoxNetworkLog{}
 
+	var destPod, destLabels, destNs string
```
nyrahul:
```suggestion
```

src/observability/summarizer.go
```diff
@@ -124,7 +130,24 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Log) {
...
+					ip = ""
+					destNs = ""
+					destLabel = ""
+					portStr = ""
+					bindPort = ""
+					bindAddress = ""
+					protocol = ""
+					nwrule = ""
+				}
+				existingClustername = syslog.ClusterName
```
nyrahul:
```suggestion
				if err == nil {
     				existingClustername = syslog.ClusterName
				}
```

src/observability/summarizer.go
```diff
@@ -124,7 +130,24 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Log) {
...
+			}
+
+			ip, destNs, destLabel, portStr, bindPort, bindAddress, protocol, nwrule, err = extractNetworkInfoFromSystemLog(*syslog, pods, services)
```
nyrahul:
```suggestion
			ip, destNs, destLabel, portStr, bindPort, bindAddress, protocol, nwrule, err := extractNetworkInfoFromSystemLog(*syslog, pods, services)
```

src/observability/summarizer.go
```diff
@@ -73,6 +73,12 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Log) {
 
 	deployments := cluster.GetDeploymentsFromK8sClient()
 
+	var ip, destLabel, destNs, portStr, bindPort, bindAddress, protocol, nwrule string
```
nyrahul:
```suggestion
```

nyrahul (changes requested):

nyrahul (approved):

PR #637, opened 2022-12-20
syscall summarised data

yasin-cs-ko-ak:
This PR Contains Summarised data of Syscalls.

This is how output will look like:
```
+----------------+-------------------------------------+--------------+-----------------------------+-------+------------------------------+--------+
| PARENT PROCESS |            CHILD PROCESS            |   SYSCALL    |         PARAMETERS          | COUNT |      LAST UPDATED TIME       | RESULT |
+----------------+-------------------------------------+--------------+-----------------------------+-------+------------------------------+--------+
| /bin/bash      | /bin/tar                            | SYS_FCHOWNAT | userid=33 group=33 mode=0   |    12 | Thu Dec  8 13:58:15 UTC 2022 | Allow |
| /bin/bash      | /bin/tar                            | SYS_FCHOWNAT | userid=33 group=33 mode=256 |     4 | Thu Dec  8 14:58:15 UTC 2022 | Allow |
| /bin/bash      | /usr/local/bin/docker-entrypoint.sh | SYS_SETUID   | userid=999                  |     7 | Thu Dec  8 13:59:15 UTC 2022 | Allow |
|                |                                     |              |                             |       |                              |        |
+----------------+-------------------------------------+--------------+-----------------------------+-------+------------------------------+--------+

```


vishnusomank (approved):
@yasin-cs-ko-ak  In the DB I can see that there are parameters without any data in them (eg: `flags=` )
Can we clean them before pushing it to the db or clean them before showing the result with `karmor summary` 
```
|default|0|0|wordpress-mysql|0|wordpress|wordpress:4.8-apache@sha256:6216f64ab88fc51d311e38c7f69ca3f9aaba621492b4f1fa93ddf63093768845|39b5eb247fb73ab63b02ad77612505018ce2c065ec2d257559c4768606d21d75|wordpress-c4bf5b44b-v7s4l|Syscall|app=wordpress|wordpress|/bin/bash|/bin/rm|||||0||||SYS_UNLINKAT|flags=|Allow|3|1671604097
````

**Comment by yasin-cs-ko-ak** [comment]
> @yasin-cs-ko-ak In the DB I can see that there are parameters without any data in them (eg: `flags=` ) Can we clean them before pushing it to the db or clean them before showing the result with `karmor summary`
> 
> ```
> |default|0|0|wordpress-mysql|0|wordpress|wordpress:4.8-apache@sha256:6216f64ab88fc51d311e38c7f69ca3f9aaba621492b4f1fa93ddf63093768845|39b5eb247fb73ab63b02ad77612505018ce2c065ec2d257559c4768606d21d75|wordpress-c4bf5b44b-v7s4l|Syscall|app=wordpress|wordpress|/bin/bash|/bin/rm|||||0||||SYS_UNLINKAT|flags=|Allow|3|1671604097
> ```

@vishnusomank Thank you for your thorough review. As discussed with @nyrahul We can leave it in parameters whatever comes after syscall. Since it will be different for every syscall so we don't have the knowledge of that. So we can keep parameters as it is. 

Do you have any other suggestions or feedback on this? @nyrahul @seswarrajan

File: src/observability/kubearmor.go [hunk 1]
```diff
@@ -280,12 +281,23 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
 				Count:       uint32(ss.Count),
 				UpdatedTime: t.Format(time.UnixDate),
 			})
+		} else if ss.Operation == "Syscall" {
```
nyrahul: constants exists for these "Syscall" and other strings ... Lets use those.

nyrahul (changes requested):

src/observability/kubearmor.go
yasin-cs-ko-ak: Thank you for your review and suggestions. There are no existing constants found. I have added 4 new constants for operations and made the changes in respective fields.

yasin-cs-ko-ak:
@vishnusomank Kindly verify latest changes. Thank you

yasin-cs-ko-ak:
Kindly review it and provide feedback.
cc: @nyrahul @vishnusomank @seswarrajan

yasin-cs-ko-ak:
Here is the issue for this PR with discussion. If anyone need further any explanation: https://github.com/kubearmor/kubearmor-client/issues/220

src/observability/summarizer.go
```diff
@@ -73,6 +73,27 @@ func extractNetworkInfoFromSystemLog(netLog pb.Log, pods []types.Pod, services [
 	return ip, destNs, destLabel, port, bindPort, bindAddress, protocol, nwrule, nil
 }
 
+func extractSyscallInfoFromSystemLog(sysCallLog pb.Log, pods []types.Pod, services []types.Service) (string, string, string, string, error) {
+	var ParentProcess, ChildProcess, Syscall, Parameters string = "", "", "", ""
+	err := errors.New("not a valid Syscall")
+
+	if sysCallLog.Operation == types.OperationTypeSyscall {
+		ParentProcess = sysCallLog.ParentProcessName
```
seswarrajan: Why are we using ParentProcessName and ProcessName instead of Source and Resource from logs.

src/observability/summarizer.go
```diff
@@ -164,6 +185,28 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Log) {
...
+					existingClustername = syslog.ClusterName
+				}
+			}
+
+			ParentProcess, ChildProcess, Syscall, Parameters, err := extractSyscallInfoFromSystemLog(*syslog, pods, services)
+
+			if err != nil {
+				continue
+			}
+			sysSummary.Source = ParentProcess
```
seswarrajan: Is it not possible to use Source and Resource from logs directly ?

src/types/constants.go
```diff
@@ -41,4 +41,10 @@ const (
 	// K8sNetworkPolicy
 	K8sNwPolicyAPIVersion = "networking.k8s.io/v1"
 	K8sNwPolicyKind       = "NetworkPolicy"
+
+	// Operation Types
+	OperationTypeFile    = "File"
```
seswarrajan:
```suggestion
	SumOpTypeFile    = "File"
```

seswarrajan (changes requested):

src/observability/summarizer.go
yasin-cs-ko-ak: We only need ParentProcessName and ProcessName for syscall and resource is not necessary.
FYI: we have edited the "ParentProcessName -> Source" when we are working "summary data".

src/observability/summarizer.go
yasin-cs-ko-ak: We only need ParentProcessName and ProcessName for syscall and resource is not necessary.
FYI: we have edited the "ParentProcessName -> Source" when we are working "summary data".

src/types/constants.go
yasin-cs-ko-ak: Thanks for the suggestions! 
Suggestions are updated.

seswarrajan (approved):
LGTM

seswarrajan (approved):

src/observability/summarizer.go
```diff
@@ -73,6 +73,29 @@ func extractNetworkInfoFromSystemLog(netLog pb.Log, pods []types.Pod, services [
...
+		if strings.Contains(resslice[0], "syscall") {
+			Syscall = strings.Split(resslice[0], "=")[1]
+		}
+		if len(resslice) >= 2 {
+			Parameters = resslice[1]
+		}
+
+	} else {
+		return "", "", "", "", err
+	}
```
nyrahul:
```suggestion
	if sysCallLog.Operation != types.SumOpTypeSyscall {
		return "", "", "", "", err
	}
	ParentProcess = sysCallLog.ParentProcessName
	ChildProcess = sysCallLog.ProcessName

	resslice := strings.SplitN(sysCallLog.Data, " ", 2)
	if strings.Contains(resslice[0], "syscall") {
		Syscall = strings.Split(resslice[0], "=")[1]
	}
	if len(resslice) >= 2 {
		Parameters = resslice[1]
	}
```

src/types/constants.go
```diff
@@ -41,4 +41,10 @@ const (
 	// K8sNetworkPolicy
 	K8sNwPolicyAPIVersion = "networking.k8s.io/v1"
 	K8sNwPolicyKind       = "NetworkPolicy"
+
+	// Operation Types
+	SumOpTypeFile    = "File"
+	SumOpTypeProcess = "Process"
+	SumOpTypeNetwork = "Network"
+	SumOpTypeSyscall = "Syscall"
```
nyrahul: Aren't these constants already defined? Maybe we need to just add `Syscall` but rest are already there.

nyrahul (changes requested):

src/types/constants.go
yasin-cs-ko-ak: Yes. Constants exists in [systempolicy](https://github.com/accuknox/discovery-engine/blob/ce809848f042feb4bd95f8afa7abd5f3e680ed4b/src/systempolicy/systemPolicy.go#L52) 

Should I import systempolicy and use those constants or leave it as it is?

src/observability/summarizer.go
yasin-cs-ko-ak: Thank you for your suggestions. Changes are commited.

src/types/constants.go
yasin-cs-ko-ak: Couldn't import systempolicy into obserability because it has "import cycle error".  We may introduce types interface and break the cycle. I guess it's not a good approach. 
If you have any thoughts on this, please let me know.

yasin-cs-ko-ak: Import error cycle error has found in systempolicy package

```
import cycle not allowed: import stack: [command-line-arguments github.com/accuknox/auto-policy-discovery/src/systempolicy github.com/accuknox/auto-policy-discovery/src/feedconsumer github.com/accuknox/auto-policy-discovery/src/plugin github.com/accuknox/auto-policy-discovery/src/observability github.com/accuknox/auto-policy-discovery/src/systempolicy]
```
Since `systempolicy` package using `plugin` package and the `plugin` has `observability`. So we couldn't `systempolicy` in `observability`.
cc: @Ankurk99

src/observability/summarizer.go
```diff
@@ -73,6 +73,27 @@ func extractNetworkInfoFromSystemLog(netLog pb.Alert, pods []types.Pod, services
 	return ip, destNs, destLabel, port, bindPort, bindAddress, protocol, nwrule, nil
 }
 
+func extractSyscallInfoFromSystemLog(sysCallLog pb.Alert, pods []types.Pod, services []types.Service) (string, string, string, string, error) {
+
+	var ParentProcess, ChildProcess, Syscall, Parameters string = "", "", "", ""
```
seswarrajan: For local variables do not start with caps. use small letter always.

seswarrajan:
```suggestion
	var parentProcess, childProcess, sysCall, parameters string = "", "", "", ""
```

src/observability/summarizer.go
```diff
@@ -165,6 +186,28 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Alert) {
 			sysSummary.Destination = strings.Split(syslog.Resource, " ")[0]
 		}
 
+		if syslog.Operation == types.SumOpTypeSyscall {
+			if existingClustername != syslog.ClusterName {
+				_, services, _, pods, err = cluster.GetAllClusterResources(syslog.ClusterName)
```
seswarrajan, line 191: I do not prefer calling GetAllClusterResources all the time. Are there any alternatives here?

seswarrajan (changes requested):

src/observability/summarizer.go
yasin-cs-ko-ak: @seswarrajan Thank you for review and suggestions. Variable names are modified.

src/observability/summarizer.go
yasin-cs-ko-ak: I have analysed the cluster package it seems there are no alternative.

src/types/constants.go
yasin-cs-ko-ak: I replaced the constants in `systempolicy` package with constants from `types/constants.go`. So for now we have system operation constants in one place.

src/observability/summarizer.go
yasin-cs-ko-ak: Thank you for your suggestion. I have modified the code.

src/observability/kubearmor.go
```diff
@@ -231,7 +232,7 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
 		Labels:        req.Label,
 	})
 	if err != nil {
-		return nil, nil, nil, types.ObsPodDetail{}
+		return nil, nil, nil, nil, types.ObsPodDetail{}
```
vishnusomank, line 226: Since we are not returning the error shouldn't we log the error here?

src/observability/kubearmor.go
```diff
@@ -278,12 +279,23 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
...
+			sysCallData = append(sysCallData, types.SysObsSycallData{
+				ParentProcess: ss.Source,
+				ChildProcess:  ss.Destination,
+				Syscall:       ss.Syscall,
+				Parameters:    ss.Parameters,
+				Count:         uint32(ss.Count),
+				UpdatedTime:   t.Format(time.UnixDate),
+				Result:        ss.Action,
+			})
 		}
```
vishnusomank, line 287: can we have an else statement here? If the received operation is not of the mentioned types we should log it. 
```
else {
   log.Debug().Msgf("Invalid Operation %s received", ss.Operation)
}

File: src/observability/kubearmor.go [hunk 11]
```diff
@@ -278,12 +279,23 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
...
+		} else if ss.Operation == types.OpTypeSyscall {
+			//ExtractsyscallData
+			sysCallData = append(sysCallData, types.SysObsSycallData{
+				ParentProcess: ss.Source,
+				ChildProcess:  ss.Destination,
+				Syscall:       ss.Syscall,
+				Parameters:    ss.Parameters,
+				Count:         uint32(ss.Count),
+				UpdatedTime:   t.Format(time.UnixDate),
+				Result:        ss.Action,
```
vishnusomank, line 283: we should rename the variable `Result` to `Status` since we are following the same for Process and File data

src/observability/summarizer.go
```diff
@@ -73,6 +73,29 @@ func extractNetworkInfoFromSystemLog(netLog pb.Alert, pods []types.Pod, services
 	return ip, destNs, destLabel, port, bindPort, bindAddress, protocol, nwrule, nil
 }
 
+func extractSyscallInfoFromSystemLog(syscallLog pb.Alert, pods []types.Pod, services []types.Service) (string, string, string, string, error) {
+	var parentProcess, childProcess, syscall, parameters string = "", "", "", ""
+	err := errors.New("not a valid Syscall")
```
vishnusomank: we do not have to create an error variable here.

src/observability/summarizer.go
```diff
@@ -73,6 +73,29 @@ func extractNetworkInfoFromSystemLog(netLog pb.Alert, pods []types.Pod, services
...
+		resslice := strings.SplitN(syscallLog.Data, " ", 2)
+		if strings.Contains(resslice[0], "syscall") {
+			syscall = strings.Split(resslice[0], "=")[1]
+		}
+		if len(resslice) >= 2 {
+			parameters = resslice[1]
+		}
+
+	} else {
+		return "", "", "", "", err
```
vishnusomank:
```suggestion
		return "", "", "", "", errors.New("not a valid syscall")
```

src/observability/summarizer.go
```diff
@@ -73,6 +73,29 @@ func extractNetworkInfoFromSystemLog(netLog pb.Alert, pods []types.Pod, services
...
+
+	if syscallLog.Operation == types.OpTypeSyscall {
+		parentProcess = syscallLog.ParentProcessName
+		childProcess = syscallLog.ProcessName
+
+		resslice := strings.SplitN(syscallLog.Data, " ", 2)
+		if strings.Contains(resslice[0], "syscall") {
+			syscall = strings.Split(resslice[0], "=")[1]
+		}
+		if len(resslice) >= 2 {
```
vishnusomank:
```suggestion
		if len(resslice) == 2 {
```

src/observability/summarizer.go
vishnusomank: since we are only using the services and pods from the  `GetAllClusterResources` function call wouldn't it be easy to just call for the specific items?

Something like the following
```
clusterInstance := cluster.GetClusterFromClusterName(syslog.ClusterName)
if clusterInstance.ClusterID != 0 {
	services = cluster.GetServicesFromCluster(clusterInstance)
	pods = cluster.GetPodsFromCluster(clusterInstance)
        existingClustername = syslog.ClusterName
}
```

@seswarrajan any suggestions?

src/observability/summarizer.go
```diff
@@ -165,6 +188,28 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Alert) {
...
+			ParentProcess, ChildProcess, Syscall, Parameters, err := extractSyscallInfoFromSystemLog(*syslog, pods, services)
+
+			if err != nil {
+				continue
+			}
+			sysSummary.Source = ParentProcess
+			sysSummary.Destination = ChildProcess
+			sysSummary.Syscall = Syscall
+			sysSummary.Parameters = Parameters
+		} else if syslog.Operation == types.OpTypeFile || syslog.Operation == types.OpTypeProcess || syslog.Operation == types.OpTypeNetwork {
```
vishnusomank: We are only populating the data if syslog.Operation is of type `syscall`. so we can set the syscall and parameters to empty in an else condition. No need for again checking the operation type.

src/observability/summary.go
```diff
@@ -12,11 +12,11 @@ func GetSummaryData(request *opb.Request) (*opb.Response, error) {
 	resp := opb.Response{}
 	var err error = nil
 
-	if strings.Contains(request.Type, "process") || strings.Contains(request.Type, "file") || strings.Contains(request.Type, "network") {
+	if strings.Contains(request.Type, "process") || strings.Contains(request.Type, "file") || strings.Contains(request.Type, "network") || strings.Contains(request.Type, "syscall") {
```
vishnusomank, line 15: can we have a switch case instead of the if conditions? I think a switch will be more appropriate since we are targeting only the `request.Type` in both conditions. @seswarrajan any issues if we use a switch case?

Something like

```
switch request.Type {
	case "process","file","network","syscall":
        // handle operation
        case "ingress","egress":
        // handle ingress and egress
        default:
        // log a debug message saying no desired request.Types were found
}
```

vishnusomank (changes requested):
@yasin-cs-ko-ak Please find comments inline

src/observability/kubearmor.go
yasin-cs-ko-ak: When we were having design conversation for syscall data we come up with `Result` instead of `Status`. Since it's a syscall it's appropriate to call Result or if you have any other thoughts let me know.

src/observability/summary.go
yasin-cs-ko-ak: Since this need more discussion on code refactoring so this can be handled separately with code refactoring PR.

src/observability/summarizer.go
yasin-cs-ko-ak: For now I am keeping it as it is. Since it's throwing error.
```sh
11:32AM ERR cluster/clusterMgmtHandler.go:88 > Error on response.
[ERROR] - Post "http://localhost:8080/cm/api/v1/cluster-management/get-cluster-ids": dial tcp [::1]:8080: connect: connection refused
```

vishnusomank: I think this is because you are not forwarding the cluster management service locally. Can you try to forward the cluster management service and test it again @yasin-cs-ko-ak

src/observability/kubearmor.go
yasin-cs-ko-ak: error message is logged here and thank you for your review.

src/observability/kubearmor.go
yasin-cs-ko-ak: The log code has been added and thank you for your review.

src/observability/summarizer.go
yasin-cs-ko-ak: updated as per your suggestions and thank you for your review.

src/observability/summarizer.go
yasin-cs-ko-ak: Added :))

src/observability/summarizer.go
yasin-cs-ko-ak: Thank you for your suggestions and the code has been modified.

src/observability/summarizer.go
yasin-cs-ko-ak: I optimised the code

src/observability/kubearmor.go
```diff
@@ -232,7 +233,8 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
 		Deployment:    req.DeployName,
 	})
 	if err != nil {
-		return nil, nil, nil, types.ObsPodDetail{}
+		log.Error().Msg(err.Error())
```
vishnusomank:
```suggestion
		log.Error().Msgf("failed to get system summary, error: %v", err)
```

yasin-cs-ko-ak: Modified! thank you

vishnusomank (approved):
LGTM!

PR #640, opened 2022-12-28
update default resource usage

Ankurk99:
Set and update default resource request and limits to reflect closer to actual usage

### DIFF
--- a/deployments/helm/values.yaml
+++ b/deployments/helm/values.yaml
@@ -72,11 +72,11 @@ volumeMountsDiscoveryEngine:
 #resource DiscoveryEngine
 resourcesDiscoveryEngine:
   requests:
-    cpu: 50m
-    memory: 20Mi
+    cpu: 100m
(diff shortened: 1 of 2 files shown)


deployments/helm/values.yaml
```diff
@@ -72,10 +72,10 @@ volumeMountsDiscoveryEngine:
 #resource DiscoveryEngine
 resourcesDiscoveryEngine:
   requests:
-    cpu: 50m
-    memory: 20Mi
+    cpu: 100m
+    memory: 100Mi
   limits:
-    cpu: 500m
+    cpu: 200m
```
nyrahul: Lets be a bit more generous with limits. 
Please set it to
```
  limits:
    cpu: 500m
    memory: 1Gi
```
Make the same change in both places.

nyrahul (changes requested):

nyrahul (approved):

PR #647, opened 2023-01-10
adding smoke tests

Prateeknandle:
Signed-off-by: Prateeknandle <prateeknandle@gmail.com>


tests/smoke/smoke_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/accuknox/auto-policy-discovery/src/types"
+	"github.com/kubearmor/KubeArmor/tests/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rs/zerolog/log"
+)
+
+var _ = BeforeSuite(func() {
+	//install wordpress-mysql app
+	err := util.K8sApply([]string{"res/wordpress-mysql-deployment.yaml"})
```
seswarrajan: My suggestion is to Deploy DE and then deploy wordpress-mysql app for test.

seswarrajan (changes requested):
1. Kindly deploy DE then deploy WP-MySQL demo app
2. Why is the go.mod file looks new. There should be an update on old one and not new one to be added. Is it that go.mod not added previously ?

tests/smoke/smoke_test.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Error().Msgf("Failed to apply the `karmor discover` command : %v", err)
+	}
+	err = json.Unmarshal(cmd, &policy)
+	if err != nil {
+		log.Error().Msgf("Failed to unmarshal the policy : %v", err)
+	}
+	return policy, err
+}
+
+var _ = Describe("Smoke", func() {
```
rksharma95, line 100: before running a test, make sure that `discovery-engine` and `wordpress-app` is up and running using `BeforeEach()`

tests/smoke/smoke_test.go
```diff
@@ -0,0 +1,70 @@
...
+	Describe("Auto Policy Discovery", func() {
+		It("test", func() {
+			policy, err := discover("wordpress-mysql", "app=wordpress")
+			Expect(err).To(BeNil())
+			Expect(policy.APIVersion).To(Equal("security.kubearmor.com/v1"))
+			Expect(policy.Kind).To(Equal("KubeArmorPolicy"))
+			Expect(policy.Metadata["namespace"]).To(Equal("wordpress-mysql"))
+			Expect(policy.Spec.Action).To(Equal("Allow"))
+			Expect(policy.Spec.Selector.MatchLabels["app"]).To(Equal("wordpress"))
+			Expect(policy.Spec.File.MatchDirectories[0].Dir).To(Equal("/usr/lib/x86_64-linux-gnu/"))
```
rksharma95: expecting data to be at fixed position i.e. at index 0 in this case, may lead to an issue and key to flakyness.

tests/smoke/smoke_test.go
Prateeknandle: go.mod was not present earlier

tests/smoke/smoke_test.go
Prateeknandle: @rksharma95 what could be a possible solution for this, storing values in map with assigning a specific key, as MatchLabels is a map?

Prateeknandle: also does these values change on different host?

rksharma95: > also can these values be changed on different host?

i don't think so, to be more confident we can target wordpress app specific policies they will not change.

> @rksharma95 what could be a possible solution for this, storing values in map with assigning a specific key, as MatchLabels is a map?

we can traverse slices i.e. MatchDirectories[ ], MatchPaths[ ] and look for the target values.

Prateeknandle: > we can traverse slices i.e. MatchDirectories[ ], MatchPaths[ ] and look for the target values.

@rksharma95 can we do something like this
```
for i := range policy.Spec.File.MatchDirectories {
				if policy.Spec.File.MatchDirectories[i].Dir == "/etc/" {
					value = "/etc/"
				}
			}
			Expect(value).To(Equal("/etc/"))
```

tests/smoke/smoke_test.go
```diff
@@ -0,0 +1,138 @@
...
+		getWpsqlPod("discovery-engine-",
+			"container.apparmor.security.beta.kubernetes.io/discovery-engine: localhost/kubearmor-accuknox-agents-discovery-engine-discovery-engine", "accuknox-agents")
+	})
+
+	AfterEach(func() {
+		util.KarmorLogStop()
+		err := util.DeleteAllKsp()
+		Expect(err).To(BeNil())
+		// wait for policy deletion
+		time.Sleep(5 * time.Second)
```
rksharma95, line 116: it's not needed here, we are not applying any security policy, 
is there any other purpose for the sleep here?

rksharma95:
test is failing when i tried it locally, @Prateeknandle can you verify if it's working?
`{"level":"error","time":"2023-01-11T10:23:56+05:30","message":"Failed to unmarshal the policy : invalid character '{' after top-level value"}
`

tests/smoke/smoke_test.go
Prateeknandle: I thought if someone had a kubearmor policy already in the cluster then it'll be removed. Should I remove this?

Prateeknandle:
> test is failing when i tried it locally, @Prateeknandle can you verify if it's working? `{"level":"error","time":"2023-01-11T10:23:56+05:30","message":"Failed to unmarshal the policy : invalid character '{' after top-level value"} `

@rksharma95 I'm not getting any error like this, I hope you're using `karmor discover -n wordpress-mysql -l "app=wordpress" -f json`, this command with json output not with yaml.

If you're using the same code then I don't understand what's the problem because I'm getting the expected result

rksharma95:
> > test is failing when i tried it locally, @Prateeknandle can you verify if it's working? `{"level":"error","time":"2023-01-11T10:23:56+05:30","message":"Failed to unmarshal the policy : invalid character '{' after top-level value"} `
> 
> @rksharma95 I'm not getting any error like this, I hope you're using `karmor discover -n wordpress-mysql -l "app=wordpress" -f json`, this command with json output not with yaml.
> 
> If you're using the same code then I don't understand what's the problem because I'm getting the expected result

i think we should have CI workflow for testing first to handle this issue better.

seswarrajan (approved):
LGTM

Prateeknandle:
will open the new pr, blundered this during rebase

Prateeknandle:
will open new pr, blundered this one during rebase

PR #649, opened 2023-01-11
Fix for invalid timestamp for Kubearmor alerts

seswarrajan:

### DIFF
--- a/src/plugin/kubearmor.go
+++ b/src/plugin/kubearmor.go
@@ -469,17 +469,21 @@ func StartKubeArmorRelay(StopChan chan struct{}, cfg types.ConfigKubeArmorRelay)
 				}
 
 				kubearmorLog := pb.Log{
-					ClusterName:   res.ClusterName,
-					ContainerName: res.ContainerName,
-					HostName:      res.HostName,
-					NamespaceName: res.NamespaceName,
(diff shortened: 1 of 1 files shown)


src/plugin/kubearmor.go
```diff
@@ -469,17 +469,21 @@ func StartKubeArmorRelay(StopChan chan struct{}, cfg types.ConfigKubeArmorRelay)
...
-					PodName:       res.PodName,
-					Source:        res.Source,
-					Operation:     res.Operation,
-					Resource:      res.Resource,
-					Data:          res.Data,
-					Result:        res.Result,
-					Type:          res.Type,
+					ClusterName:       res.ClusterName,
+					ContainerName:     res.ContainerName,
+					ContainerID:       res.ContainerID,
```
nyrahul, line 474: Cant we directly assign `res` to `kubearmorLog`? Like what we do in the case of telemetry events above.

nyrahul (changes requested):

src/plugin/kubearmor.go
nyrahul: We need policy-name to be part of this.

nyrahul (approved):

PR #657, opened 2023-01-30
remove process fromsource

achrefbensaad:
remove fromsource from process

Signed-off-by: Achref ben saad <achref@accuknox.com>


seswarrajan (approved):
The changes LGTM. But what are we achieving from this change @achrefbensaad .

seswarrajan:
The changes LGTM. But what are we achieving from this change @achrefbensaad .

nyrahul (approved):

PR #658, opened 2023-01-30
Recommend host hardening policies

vishnusomank:
- show KubearmorHostPolicies while recommending hardening policies

Signed-off-by: Vishnu Soman <vishnu@accuknox.com>

### DIFF
--- a/deployments/k8s/deployment.yaml
+++ b/deployments/k8s/deployment.yaml
@@ -157,13 +157,22 @@ metadata:
   namespace: accuknox-agents
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: discovery-engine-role
+rules:
+- apiGroups: ["*"]
+  resources: ["pods", "services", "deployments", "endpoints", "namespaces", "nodes"]
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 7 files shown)


Ankurk99 (approved):
I have verified that with these roles DE is running fine.

deployments/k8s/deployment.yaml
```diff
@@ -157,13 +157,22 @@ metadata:
   namespace: accuknox-agents
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: discovery-engine-role
+rules:
```
seswarrajan, line 163: As per my understanding, this changes involve removing DE to run in admin mode/role and make it possible to run in non-admin role with identified permissions. 

Kindly handle this in a separate PR as this will be easy to track.

seswarrajan (changes requested):

deployments/k8s/deployment.yaml
Ankurk99: Would agree :+1:

Ankurk99:
@vishnusomank can you please open a new PR for the privileged role changes?

vishnusomank:
@Ankurk99 yes. i'll close this PR and re-raise it

Ankurk99:
No, you could have just removed that particular commit and cherry picked it to another PR :smile:

PR #664, opened 2023-02-07
Fix gosec issue

Prateeknandle:


Prateeknandle:
@rksharma95 please review the pr

src/cluster/clusterMgmtHandler.go
```diff
@@ -89,7 +89,11 @@ func getResponseBytes(method string, url string, data map[string]interface{}) []
 		return nil
 	}
 	dumpHttpClient(nil, resp)
-	defer resp.Body.Close()
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Error().Msgf("Error closing http stream %s\n", err)
```
rksharma95: I'm not sure if log at error level is needed here. ig it's recoverable in subsequent requests. imo we can use warn here, WDYT?

Prateeknandle: yes ig you're right, will update

rksharma95 (approved):
LGTM :+1: please squash your commits and it's ready to merge

PR #672, opened 2023-02-15
[Summary data] Removing `AF_UNIX `and `AF_NETLINK` (showing only  `AF_INET `and `AF_INET6`)

### LINKED ISSUE #671: removing `AF_NETLINK` and `AF_UNIX` from summary data
## Reasons to remove these protocols.
(This is related to karmor summary network data bind points)
- Both protocols has no significance from security perspective. 
- It is unnecessarily complicating the app behaviour summary.

## Solution
- Only showing `AF_INET` and `AF_INET6` which is equivalent to **IPv4** and **IPv6**

yasin-cs-ko-ak:
This PR fixes this issue: #671

### DIFF
--- a/src/observability/summarizer.go
+++ b/src/observability/summarizer.go
@@ -40,14 +40,13 @@ func extractNetworkInfoFromSystemLog(netLog pb.Alert, pods []types.Pod, services
 		}
 	} else if strings.Contains(netLog.Resource, "AF_UNIX") {
 		var path string
-		protocol = "AF_UNIX"
 		resslice := strings.Split(netLog.Resource, " ")
 		for _, locres := range resslice {
 			if strings.Contains(locres, "sun_path") {
 				path = strings.Split(locres, "=")[1]
 				if path != "" {
 					ip = path
-					bindAddress = path
+					protocol = "AF_UNIX"
 					break
 				}
 			}
--- a/src/observability/summary.go
+++ b/src/observability/summary.go
@@ -81,17 +81,19 @@ func GetSummaryData(request *opb.Request) (*opb.Response, error) {
 						UpdatedTime: loc_nw.UpdatedTime,
 					})
 				} else if loc_nw.NetType == "bind" {
-					bindNwResp = append(bindNwResp, &opb.SysNwSummaryData{
-						Protocol:    loc_nw.Protocol,
-						Command:     loc_nw.Command,
-						IP:          loc_nw.PodSvcIP,
-						BindPort:    loc_nw.BindPort,
-						BindAddress: loc_nw.BindAddress,
-						Labels:      loc_nw.Labels,
-						Namespace:   loc_nw.Namespace,
-						Count:       strconv.Itoa(int(loc_nw.Count)),
-						UpdatedTime: loc_nw.UpdatedTime,
-					})
+					if strings.EqualFold(loc_nw.Protocol, "AF_INET6") || strings.EqualFold(loc_nw.Protocol, "AF_INET6") {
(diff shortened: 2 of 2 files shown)


src/observability/summarizer.go
```diff
@@ -40,14 +40,13 @@ func extractNetworkInfoFromSystemLog(netLog pb.Alert, pods []types.Pod, services
...
 		var path string
-		protocol = "AF_UNIX"
 		resslice := strings.Split(netLog.Resource, " ")
 		for _, locres := range resslice {
 			if strings.Contains(locres, "sun_path") {
 				path = strings.Split(locres, "=")[1]
 				if path != "" {
 					ip = path
-					bindAddress = path
+					protocol = "AF_UNIX"
```
nyrahul, line 49: @seswarrajan , do you understand why this change?

seswarrajan: No @nyrahul 
I've to skim through the code.

seswarrajan, line 49: @yasin-cs-ko-ak This does not seem like a fix. We need to fix/remove the code carefully so that AF_NETLINK/AF_UNIX shows for proper values.

seswarrajan (changes requested):

src/observability/summarizer.go
yasin-cs-ko-ak: @seswarrajan that value only shows for ingress/egress. In bind it is only AF_INET and AF_INET6. 
or we can change the AF_UNIX to UNIX as it is before.

yasin-cs-ko-ak: that AF_UNIX is not interfere with bind anymore.

seswarrajan: I understand that AF_UNIX and AF_NETLINK should not be present in the code itself and needs to be removed. 
We can discuss once.

seswarrajan (changes requested):

src/observability/summary.go
```diff
@@ -81,17 +81,19 @@ func GetSummaryData(request *opb.Request) (*opb.Response, error) {
...
-						Command:     loc_nw.Command,
-						IP:          loc_nw.PodSvcIP,
-						BindPort:    loc_nw.BindPort,
-						BindAddress: loc_nw.BindAddress,
-						Labels:      loc_nw.Labels,
-						Namespace:   loc_nw.Namespace,
-						Count:       strconv.Itoa(int(loc_nw.Count)),
-						UpdatedTime: loc_nw.UpdatedTime,
-					})
+					if strings.EqualFold(loc_nw.Protocol, "AF_INET6") || strings.EqualFold(loc_nw.Protocol, "AF_INET6") {
```
yasin-cs-ko-ak, line 84: @seswarrajan That's why I added here only to show inet and inet6. 

[L84](https://github.com/accuknox/discovery-engine/blob/3b5942ab845beec9ac9af0d526583c17e29aebe2/src/observability/summary.go#L84)

yasin-cs-ko-ak:
@seswarrajan Kindly review it and approve it. If there's anything to discuss more, kindly let me know. Thanks

seswarrajan:
@yasin-cs-ko-ak When this data is found in DE code, this needs to be completely ignored from processing. 
Not to be handled at all.

yasin-cs-ko-ak:
> @yasin-cs-ko-ak When this data is found in DE code, this needs to be completely ignored from processing. 
> Not to be handled at all.

@seswarrajan if we remove this completely it won't shown for egress and ingress. We are only removing it for bind only right?
Any thoughts on that?

seswarrajan:
Yes.. Bind only..

PR #674, opened 2023-02-22
removing namespace dependency of kubearmor for discovery-engine

### LINKED ISSUE #660: Remove namespace dependency for KubeArmor
**Description**
Currently Discovery Engine configuration checks for KubeArmor in the default kube-system namespace only. If KubeArmor is installed in any other namespace then it won't be able to communicate with that. 
It causes a problem that we can't really install KubeArmor in other namespace and use discovery-engine with it. 

**Solution**
The aim is to remove this the dependency on kube-system namespace and use another one (for eg. `kubearmor` namespace). Then we will be installing KubeArmor in `kubearmor` namespace by default.

Prateeknandle:
closes #660 

after this discovery-engine will be able to detect kubearmor from any namespace, right noe it only detect kubearmor if it is deployed in `kube-system` namespace.

changes:
- get namespace in which kubearmor is deployed
- if kubearmor is not found in `kube-system`, then we will get the namespace in which it is deployed and then connect discovery-engine with new configurations.

### DIFF
--- a/src/cluster/k8sClientHandler.go
+++ b/src/cluster/k8sClientHandler.go
@@ -459,3 +459,28 @@ func GetDeploymentsFromK8sClient() []types.Deployment {
 	}
 	return results
 }
+
+func GetKubearmorRelayURL() string {
+	var namespace string
+	client := ConnectK8sClient()
+	if client == nil {
+		return ""
+	}
+
+	// get pods from k8s api client
+	pods, err := client.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{
+		LabelSelector: "kubearmor-app",
+	})
+	if err != nil {
+		log.Error().Msg(err.Error())
+		return ""
+	}
+	// for _, pod := range pods.Items {
+	// 	if pod.Labels["kubearmor-app"] == "kubearmor" {
+	// 		namespace = pod.Namespace
+	// 	}
+	// }
+	namespace = pods.Items[0].Namespace
+	url := "kubearmor." + namespace + ".svc.cluster.local"
+	return url
+}
--- a/src/cluster/k8sClientHandler.go
+++ b/src/cluster/k8sClientHandler.go
@@ -467,19 +467,17 @@ func GetKubearmorRelayURL() string {
 		return ""
 	}
 
-	// get pods from k8s api client
+	// get kubearmor-relay pod from k8s api client
 	pods, err := client.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{
-		LabelSelector: "kubearmor-app",
+		LabelSelector: "kubearmor-app=kubearmor-relay",
 	})
 	if err != nil {
 		log.Error().Msg(err.Error())
 		return ""
 	}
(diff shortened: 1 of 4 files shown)


src/cluster/k8sClientHandler.go
```diff
@@ -475,3 +475,25 @@ func GetNodesFromK8sClient() (*v1.NodeList, error) {
...
+	var namespace string
+	client := ConnectK8sClient()
+	if client == nil {
+		return ""
+	}
+
+	// get kubearmor pods from k8s api client
+	pods, err := client.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{
+		LabelSelector: "kubearmor-app",
+	})
```
daemon1024: Let's directly fetch the relay pod. 

Ref https://github.com/kubearmor/KubeArmor/blob/d6db731c7e4383faa7cd0518b750703dc017e71e/deployments/get/objects.go#L110

```suggestion
	pods, err := client.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{
		LabelSelector: "kubearmor-app=kubearmor-relay",
	})
```

src/cluster/k8sClientHandler.go
```diff
@@ -475,3 +475,25 @@ func GetNodesFromK8sClient() (*v1.NodeList, error) {
...
+	if err != nil {
+		log.Error().Msg(err.Error())
+		return ""
+	}
+	namespace = pods.Items[0].Namespace
+	url := "kubearmor." + namespace + ".svc.cluster.local"
```
daemon1024, line 498: We need to return empty if the Pod List is  empty array as well,  which suggests there's no Kubearmor deployment active.

src/cluster/k8sClientHandler.go
```diff
@@ -475,3 +475,25 @@ func GetNodesFromK8sClient() (*v1.NodeList, error) {
 	}
 	return nodeList, nil
 }
+
+func GetKubearmorRelayURL() string {
+	var namespace string
+	client := ConnectK8sClient()
+	if client == nil {
+		return ""
+	}
```
daemon1024, line 484: Is there no global client which we can access in the engine?  Do we create a new client every time we make k8s API calls in the engine?

daemon1024 (changes requested):
Please find review comments inline.

src/cluster/k8sClientHandler.go
Prateeknandle: yes, right now there is no global client present, therefore we need to create a new client every time

daemon1024: Let's create a note for it, might  need to handle it in future.

daemon1024 (approved):

nyrahul (approved):

Ankurk99:
@Prateeknandle Have we verified these changes? Like if `karmor summary` and `karmor discover` works with this change?
Because I tried installing KubeArmor in a different namespace and Discovery Engine is not able to show observability or discover policy. 
@Prateeknandle Can you please verify this?

PR #679, opened 2023-03-06
Issue #675: karmor summary flags are ignored

### LINKED ISSUE #675: `karmor summary` flags are ignored
https://github.com/accuknox/discovery-engine/blob/4117eadb3716ec2c88acbbc94a57495f44c63ce5/src/observability/kubearmor.go#L219-L227

Only `req.PodName` is respected, the rest of the parameters in the request are ignored.

```bash
$ karmor summary -p pod-1 --container redis
```
```
local port to be used for port forwarding discovery-engine-746fccb855-nlr6k: 32897 

  Pod Name        pod-1  
  Namespace Name  default                             
  Cluster Name    default                             
  Container Name  nginx                               
  Labels          app=nginx-client-side               
...
```

Vyom-Yadav:
Closes #675 

PR for kubearmor-client: https://github.com/kubearmor/kubearmor-client/pull/276

### DIFF
--- a/src/observability/kubearmor.go
+++ b/src/observability/kubearmor.go
@@ -223,7 +223,11 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
 	var podInfo types.ObsPodDetail
 
 	sysSummary, err := libs.GetSystemSummary(CfgDB, types.SystemSummary{
-		PodName: req.PodName,
+		PodName:       req.PodName,
+		NamespaceName: req.NameSpace,
+		ContainerName: req.ContainerName,
+		ClusterName:   req.ClusterName,
+		Labels:        req.Label,
 	})
 	if err != nil {
 		return nil, nil, nil, types.ObsPodDetail{}


seswarrajan (commented):
@Vyom-Yadav 
This change LGTM. 

Is it possible to modify the protobug request structure to include deployment_name as well, The SQL code already supports the change for that.

Vyom-Yadav:
> Is it possible to modify the protobug request structure to include deployment_name as well, The SQL code already supports the change for that.

Do we want deployment_name in response too?

seswarrajan:
It would be better if we have the deployment name in field as well.

Vyom-Yadav:
We return just the pod_name in response, do we want to return the deployment_name too? Of course, it will be shown only if it's not empty.

seswarrajan (approved):

seswarrajan:
> We return just the pod_name in response, do we want to return the deployment_name too? Of course, it will be shown only if it's not empty.

We can handle this in a separate PR

PR #681, opened 2023-03-09
get hardening policies for `replicasets`, `statefulsets` & `daemonsets`

### LINKED ISSUE #670: Hardening policies only getting generated for deployments
### Problem Statement:
The discovery engine is not able to recommend hardening policies if the pods are generated through a statefulset or a replicaset. With the current implementation the discovery engine checks for all deployments in the Kubernetes cluster and generates the recommended/hardening policies. 

### Steps to reproduce the issue:

1. Create a [k3s cluster](https://github.com/kubearmor/KubeArmor/tree/main/contribution/k3s)
2. Install karmor cli 
    ```
    curl -sfL http://get.kubearmor.io/ | sudo sh -s -- -b /usr/local/bin
    ```
3. Install kubearmor
    ```
    karmor install
    ```
4. Deploy `frontend` pod using replicaset
    ```
    kubectl create ns rs-demo
    echo "
    apiVersion: apps/v1
    kind: ReplicaSet
    metadata:
      name: frontend
      labels:
        app: guestbook
        tier: frontend
    spec:
      # modify replicas according to your case
      replicas: 3
      selector:
        matchLabels:
          tier: frontend
      template:
        metadata:
          labels:
            tier: frontend
        spec:
          containers:
          - name: php-redis
            image: gcr.io/google_samples/gb-frontend:v3 " | kubectl -n rs-demo apply -f- 
    ```
5. use karmor recommend to generate hardening policy
    ```
    karmor recommend -n rs-demo
    ```

The default output folder `out` will be empty, meaning no policies generated for the pods

Prateeknandle:
fixes #670 

changes:
- getting `replicasets`, `daemonsets` and `statefulsets` from k8s client and generating hardening policies respectively
- updating `clusterRole` for the same
- added test cases to test whether we get the hardening policy or not


vishnusomank:
LGTM! 

@Prateeknandle  Can we update the helm chart as well to match the `ClusterRole` ?

Prateeknandle:
> @Prateeknandle Can we update the helm chart as well to match the `ClusterRole` ?

updated

seswarrajan (approved):
LGTM

Prateeknandle:
can we merge the pr?

nyrahul:
> can we merge the pr?

The test is failing!

nyrahul:
Can we add a testcase to test this?

Prateeknandle:
> Can we add a testcase to test this?

tests for `recommend` are defined in `kubearmor-client`

> The test is failing!

it's because the tests are flaky

For testing whether changes are working or not we can see logs

vishnusomank (approved):
LGTM! 

Tested by deploying via deployment.yaml and helm. The code changes are able to detect and generate policies for Statefulset and Replicaset. 
Only issue is if a new replicaset or statefulset is deployed we'll have to wait for the cronjob to trigger (every 1 hour) to get hardening policies for those.

Prateeknandle:
With the new changes I've checked locally, tests are passing and discovery-engine is able to get the hardening policies.

the ginkgo test which are failing are the flaky ones and not related to this pr.

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,46 @@ func RecommendPolicyMain() {
...
 
-		for _, ns := range nsNotFilter {
+		for _, ns := range NsNotFilter {
 			if d.Namespace != ns {
-				generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
+				policy = generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
 			}
 		}
 	}
+	for _, pol := range policy {
```
vishnusomank: we don't need to loop through policy to append it. We can use something like
```
policies = append(policies, policy...)
```

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,46 @@ func RecommendPolicyMain() {
...
+		policies = append(policies, pol)
+	}
+	for _, r := range replicaSets.Items {
+		for _, ns := range NsNotFilter {
+			if r.Namespace != ns {
+				policy = generateHardenPolicy(r.Name, r.Namespace, r.Spec.Template.Labels)
+			}
+		}
+	}
+	for _, pol := range policy {
```
vishnusomank: same as the above comment

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,46 @@ func RecommendPolicyMain() {
...
+		policies = append(policies, pol)
+	}
+	for _, r := range statefulSets.Items {
+		for _, ns := range NsNotFilter {
+			if r.Namespace != ns {
+				policy = generateHardenPolicy(r.Name, r.Namespace, r.Spec.Template.Labels)
+			}
+		}
+	}
+	for _, pol := range policy {
```
vishnusomank: same as the above comment

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,46 @@ func RecommendPolicyMain() {
...
+		policies = append(policies, pol)
```
vishnusomank: We are not using this `policies` anywhere after it is returned right? Do we need this policies for any operation?

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,46 @@ func RecommendPolicyMain() {
...
+func generateHardenPolicy(name, namespace string, labels LabelMap) []types.KnoxSystemPolicy {
+	log.Info().Msgf("Generating hardening policy for: %v in namespace: %v", name, namespace)
 	policies, err := generatePolicy(name, namespace, labels)
 	if err != nil {
 		log.Error().Msg(err.Error())
-		return
+		return nil
 	}
 	systempolicy.UpdateSysPolicies(policies)
+	return policies
```
vishnusomank, line 191: If we are returning the policies array from this function we can remove the `systempolicy.UpdateSysPolicies(policies)` from this function and use it in `RecommendPolicyMain()` so that it'll only get called once after we have all the policies for deployment, statefulset and replicasets.

src/recommendpolicy/recommendPolicy.go
Prateeknandle: We are using [GetHardenPolicy()](https://github.com/accuknox/discovery-engine/pull/681/files#diff-507e0011278270a66cc4717d5f666421567b9ca156b62c4c01670f52e6867a44R225) to get the policies for testing, RecommendPolicyMain() will directly not return it therefore using this func.

src/recommendpolicy/recommendPolicy.go
Prateeknandle: updated please check

src/recommendpolicy/recommendPolicy.go
```diff
@@ -135,6 +142,16 @@ func RecommendPolicyMain() {
 		log.Error().Msg(err.Error())
 		return
 	}
+	replicaSets, err := client.AppsV1().ReplicaSets("").List(context.Background(), metav1.ListOptions{})
```
vishnusomank, line 133: @Prateeknandle When we get all the replicasets there is a possibility that the replicaset was created by a deployment. If that is the case then we will receive duplicate policies for the same. 
For eg., If we deploy wordpress-mysql application, we can see that it'll create replicasets from the deployment. 
```
k -n wordpress-mysql get deploy,rs                                                                                  (get_hardening_pol↓1|✚1)
NAME                        READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/mysql       0/1     1            0           6d3h
deployment.apps/wordpress   1/1     1            1           6d3h

NAME                                   DESIRED   CURRENT   READY   AGE
replicaset.apps/mysql-76ddc6ddc4       1         1         0       6d3h
replicaset.apps/wordpress-787f45786f   1         1         1       6d3h
```
Naturally we wouldn't want policies to be generated for both (ie, duplicate) Can we handle this scenario as well?

tests/recommend/recommned_test.go
```diff
@@ -0,0 +1,85 @@
...
+			flag := false
+			recommendpolicy.NsNotFilter = []string{"kube-system"}
+			policy := recommendpolicy.GetHardenPolicy()
+			Expect(policy).NotTo(BeNil())
+			Expect(len(policy)).NotTo(Equal(0))
+			for _, p := range policy {
+				if p.Metadata["namespace"] == "rs-demo" && p.Spec.Selector.MatchLabels["app"] == "nginx" {
+					flag = true
+				}
+				if flag == true {
```
vishnusomank: we can use the break in the first if condition itself. 
```
if p.Metadata["namespace"] == "rs-demo" && p.Spec.Selector.MatchLabels["app"] == "nginx" {
    flag = true
    break
}
```

src/recommendpolicy/recommendPolicy.go
Prateeknandle: Handled in new commit

tests/recommend/recommned_test.go
Prateeknandle: Handled in new commit

tests/recommend/recommned_test.go
```diff
@@ -0,0 +1,81 @@
...
+	// check for statefulset pod
+	checkPod("my-statefulset-",
+		"container.apparmor.security.beta.kubernetes.io/my-container: localhost/kubearmor-default-my-statefulset-my-container", "default")
+})
+
+var _ = AfterSuite(func() {
+
+	// create new namespace
+	_, err := util.Kubectl(fmt.Sprintf("delete namespace rs-demo"))
+	Expect(err).To(BeNil())
```
vishnusomank, line 79: can we have both the sts and rs on the same namespace and make sure that they are deleted after the test is finished?
Currently the sts is not getting deleted after test finishes. @Prateeknandle

vishnusomank (approved):

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,40 @@ func RecommendPolicyMain() {
...
-				generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
+				policy = generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
 			}
 		}
 	}
+	policies = append(policies, policy...)
+	for _, r := range replicaSets.Items {
+		for _, ns := range NsNotFilter {
+			if r.Namespace != ns && len(r.ObjectMeta.OwnerReferences) == 0 {
+				policy = generateHardenPolicy(r.Name, r.Namespace, r.Spec.Template.Labels)
```
nyrahul: Assume there are multiple namespaces with statefulset. When this for loop ends it will contain the harden `policy` only for the last namespace. All other namespaces hardening policies will be lost/overwritten. Am I missing something here?

nyrahul (changes requested):

src/recommendpolicy/recommendPolicy.go
```diff
@@ -135,6 +142,16 @@ func RecommendPolicyMain() {
 		log.Error().Msg(err.Error())
 		return
 	}
+	replicaSets, err := client.AppsV1().ReplicaSets("").List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		log.Error().Msg(err.Error())
```
nyrahul:
```suggestion
		log.Error().Msg("error getting replicasets err=" + err.Error())
```

src/recommendpolicy/recommendPolicy.go
```diff
@@ -135,6 +142,16 @@ func RecommendPolicyMain() {
...
+		return
+	}
+	statefulSets, err := client.AppsV1().StatefulSets("").List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		log.Error().Msg(err.Error())
```
nyrahul:
```suggestion
		log.Error().Msg("error getting statefulsets err=" + err.Error())
```

nyrahul (changes requested):

src/recommendpolicy/recommendPolicy.go
Prateeknandle: thanks for looking into it, updated the pr, please check.

src/recommendpolicy/recommendPolicy.go
Prateeknandle: these are also addressed

src/recommendpolicy/recommendPolicy.go
```diff
@@ -143,22 +160,42 @@ func RecommendPolicyMain() {
 			DeployNsName = append(DeployNsName, *deploy)
 		}
 
-		for _, ns := range nsNotFilter {
+		for _, ns := range NsNotFilter {
 			if d.Namespace != ns {
-				generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
+				policy = generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
```
vishnusomank: since we are not returning any error we can directly append the return value to `policies` like:
```
policies = append(policies, generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)...)
```

Prateeknandle: updated

src/recommendpolicy/recommendPolicy.go
```diff
@@ -52,6 +52,11 @@ type LabelMap = map[string]string
 // DeployNsName stores the identified deployments in a namespace
 var DeployNsName []types.Deployment
 
+// NsNotFilter stores the namespace which needs to be filtered out
+var NsNotFilter = cfg.CurrentCfg.ConfigSysPolicy.NsNotFilter
+
+var policies []types.KnoxSystemPolicy
```
nyrahul: This is a very bad practice to keep the variables global. The functions using `policies` can no more be used in multi-threaded envs.

`policies` needs to be local vars.

nyrahul (changes requested):

src/recommendpolicy/recommendPolicy.go
Prateeknandle: handled the bad func, PTAL

src/recommendpolicy/recommendPolicy.go
```diff
@@ -95,6 +101,7 @@ func StopRecommendWorker() {
 func StartRecommendCronJob() {
 	// init cron job
 	RecommendCronJob = cron.New()
+	NsNotFilter = cfg.CurrentCfg.ConfigSysPolicy.NsNotFilter
```
nyrahul: Why is this done multiple times? I see that NsNotFilter is already initialized at line 78.

src/recommendpolicy/recommendPolicy.go
```diff
@@ -122,43 +129,37 @@ func StopRecommendCronJob() {
 // RecommendPolicyMain generates recommended policies from policy-template GH
 func RecommendPolicyMain() {
 
-	nsNotFilter := cfg.CurrentCfg.ConfigSysPolicy.NsNotFilter
```
vishnusomank: can we have the `nsNotFilter` string created in this function as a local variable and pass it to `GetHardenPolicy()` rather than having it as a Global Variable

vishnusomank (changes requested):

src/recommendpolicy/recommendPolicy.go
Prateeknandle: added

src/recommendpolicy/recommendPolicy.go
```diff
@@ -180,3 +176,43 @@ func uniqueNsDeploy(deployName, deployNamespace string) *types.Deployment {
 
 	return &deploy
 }
+
+func GetHardenPolicy(deployments *v1.DeploymentList, replicaSets *v1.ReplicaSetList, statefulSets *v1.StatefulSetList, nsNotFilter []string) []types.KnoxSystemPolicy {
+
+	var policies []types.KnoxSystemPolicy
+	if !isLatest() {
+		if _, err := DownloadAndUnzipRelease(); err != nil {
+			log.Error().Msgf("Unable to download %v", err.Error())
```
nyrahul, line 229: Should this continue if the operation fails?

nyrahul: It would be good to print the version returned by `DownloadAndUnzipRelease` so as to track.

src/recommendpolicy/recommendPolicy.go
```diff
@@ -180,3 +176,43 @@ func uniqueNsDeploy(deployName, deployNamespace string) *types.Deployment {
...
+		for _, ns := range nsNotFilter {
+			if d.Namespace != ns {
+				policies = append(policies, generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)...)
+			}
+		}
+	}
+
+	for _, r := range replicaSets.Items {
+		for _, ns := range nsNotFilter {
+			if r.Namespace != ns && len(r.ObjectMeta.OwnerReferences) == 0 {
```
nyrahul, line 250: Why is the check `len(r.ObjectMeta.OwnerReferences) == 0` not present for statefulsets/deployments?

nyrahul (changes requested):

src/recommendpolicy/recommendPolicy.go
Prateeknandle: this was added in order to handle [comment](https://github.com/accuknox/discovery-engine/pull/681#pullrequestreview-1352921553), this case will not occur for sts and deployments.

Ankurk99:
Hey @Prateeknandle, Can you please resolve the conflicts in the PR?

src/recommendpolicy/recommendPolicy.go
Ankurk99: It can be possible that deployment are managed by `job` owner. I have added the checks for those cases.

vishnusomank (commented):
I think we should handle daemonsets as well in this. @Ankurk99 what do you think?

src/recommendpolicy/recommendPolicy.go
```diff
@@ -199,3 +219,55 @@ func uniqueNsDeploy(deployName, deployNamespace string) *types.Deployment {
...
+
+	var policies []types.KnoxSystemPolicy
+	if !isLatest() {
+		version, err := DownloadAndUnzipRelease()
+		if err != nil {
+			log.Error().Msgf("Unable to download %v", err.Error())
+			return nil
+		} else {
+			log.Info().Msgf("Downloaded version: %v", version)
+		}
```
nyrahul:
```suggestion
			return nil
		}
		log.Info().Msgf("Downloaded version: %v", version)
```

seswarrajan (approved):
LGTM

nyrahul (approved):
Mostly good. Just one small fix.

nyrahul (approved):

nyrahul:
@vishnusomank , can you provide your approval?

vishnusomank (approved):

src/recommendpolicy/recommendPolicy.go
```diff
@@ -144,7 +162,7 @@ func RecommendPolicyMain() {
 			DeployNsName = append(DeployNsName, *deploy)
 		}
 
-		for _, ns := range nsNotFilterSysPolicy {
+		for _, ns := range nsNotFilter {
 			if d.Namespace != ns {
 				generateHardenPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
```
Vyom-Yadav, line 167: We are already generating policies for `Deployment`s here. Why are we generating them again at line 150? It should be in a single place only.

src/recommendpolicy/recommendPolicy.go
```diff
@@ -199,3 +219,54 @@ func uniqueNsDeploy(deployName, deployNamespace string) *types.Deployment {
...
+
+	var policies []types.KnoxSystemPolicy
+	if !isLatest() {
+		version, err := DownloadAndUnzipRelease()
+		if err != nil {
+			log.Error().Msgf("Unable to download %v", err.Error())
+			return nil
+		}
+		log.Info().Msgf("Downloaded version: %v", version)
+	}
```
Vyom-Yadav, line 233: This should not be clubbed with `GetHardenPolicy(...)`. Downloading and unzipping release is a separate process. Policy templates are required for both KubeArmor and Kyverno policies.

PR #690, opened 2023-03-15
Workload Attestation | MTLS Connection

PaveenV:
1. Through workloadAPI in GetSource function getting source(svid) from Spire Agent(agent-operator).
2. Passing this source to the GRPC server configuration to accept MTLS connections only.
3. Added Spire Configrations in the config file.


wazir-ahmed:
DE is not communicating with the SaaS directly. It only communicate with FS. So these changes are not needed. Closing this PR.

PR #697, opened 2023-03-23
Display summary data per deployment

### LINKED ISSUE #696: Display summary data per deployment
Current implementation of discovery engine generates summary data per pod. If there are multiple pods for a single deployment/replicasets then the summary will be lengthy. To reduce the noise and show a clean summary data,  discovery engine should generate the data per deployment. This will make sure that all the data from pods of deployment/statefulset will be merged under one table and shown.

vishnusomank:
- Modified the observability to include Summary data per deployment
- Added deployment name to struct PodInfo
- Modified codebase to get details of replicasets and statefulset along with deployments
- Modified codebase to include deployment name as part of PodInfo

Fixes: https://github.com/accuknox/discovery-engine/issues/696

### Related PR's
1. [kubearmor-client #299](https://github.com/kubearmor/kubearmor-client/pull/299)


src/cluster/k8sClientHandler.go
seswarrajan, line 1: Kindly handle lint errors.

Vyom-Yadav:
@vishnusomank Can you share the output as well?

Also, the summary data may vary per pod if someone exec'd into that particular pod, is that something we should consider?

vishnusomank:
> @vishnusomank Can you share the output as well?
> 
> Also, the summary data may vary per pod if someone exec'd into that particular pod, is that something we should consider?

@Vyom-Yadav there is no change in the codebase we are only sending the data arranged by deployment name instead of pod name. So internally the storage and function calls are almost the same.

You can refer this PR for output https://github.com/kubearmor/kubearmor-client/pull/299

src/observability/summarizer.go
```diff
@@ -127,7 +127,7 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Alert) {
 		sysSummary.Deployment = ""
 
 		for _, d := range deployments {
-			if d.Labels == syslog.Labels && d.Namespace == syslog.NamespaceName {
+			if strings.Contains(syslog.Labels, d.Labels) && d.Namespace == syslog.NamespaceName {
```
seswarrajan, line 130: I need clarification in this logic @vishnusomank 
Why are we using strings.Contains instead of == ?

vishnusomank: @seswarrajan  The `syslog.Labels` sometimes contains multiple labels as a single string and when we use == to compare it with `d.Labels` it fails. So to reduce the failure rate I added the `strings.Contains` condition.

eg: 
syslog.Labels 
 ```
sidecar.istio.io/inject=false,app=istiod,install.operator.istio.io/owning-resource=unknown,istio=pilot istio.io/rev=default,operator.istio.io/component=Pilot
```
d.Labels
```
app=istiod
```

In this scenario, the pod belongs to the deployment `istiod` but if we use == it'll not be able to register.

File: src/cluster/k8sClientHandler.go [hunk 2]
```diff
@@ -458,6 +457,92 @@ func GetDeploymentsFromK8sClient() []types.Deployment {
...
+	for _, rs := range replicasets.Items {
+		if rs.OwnerReferences == nil {
+			if rs.Namespace == "kube-system" {
+				continue
+			}
+
+			var label string
+
+			for k, v := range rs.Spec.Selector.MatchLabels {
+				label = k + "=" + v
```
Ankurk99: Do we need to handle the condition when label is empty?

Ankurk99: With multiple MatchLabels, the label variable will be overwritten (inside for loop) and only the last key-value pair in MatchLabels will be stored in it?

src/cluster/k8sClientHandler.go
```diff
@@ -458,6 +457,92 @@ func GetDeploymentsFromK8sClient() []types.Deployment {
...
+	for _, sts := range statefulset.Items {
+		if sts.OwnerReferences == nil {
+			if sts.Namespace == "kube-system" {
+				continue
+			}
+
+			var label string
+
+			for k, v := range sts.Spec.Selector.MatchLabels {
+				label = k + "=" + v
```
Ankurk99: Same as above. How are we using the multiple `label` values?

Ankurk99 (changes requested):
Please find comments inline

src/cluster/k8sClientHandler.go
```diff
@@ -446,17 +445,112 @@ func GetDeploymentsFromK8sClient() []types.Deployment {
...
 
-		var label string
+		if d.Spec.Selector.MatchLabels != nil {
+			var labels []string
 
-		for k, v := range d.Spec.Selector.MatchLabels {
-			label = k + "=" + v
+			for k, v := range d.Spec.Selector.MatchLabels {
+				label := k + "=" + v
+				labels = append(labels, label)
```
Ankurk99: what about using `labels = append(labels, k+"="+v)` ?

src/cluster/k8sClientHandler.go
```diff
@@ -446,17 +445,112 @@ func GetDeploymentsFromK8sClient() []types.Deployment {
...
+// ================= //
+// == ReplicaSet == //
+// ================= //
+
+func GetReplicaSetsFromK8sClient() []types.Deployment {
+	results := []types.Deployment{}
+
+	client := ConnectK8sClient()
+	if client == nil {
+		return results
```
Ankurk99, line 489: Should we log error or handle the case when `results` is empty?

seswarrajan (approved):
LGTM

Ankurk99 (approved):
Please handle the go-lint warnings, rest LGTM

PR #701, opened 2023-03-31
feat: Add logic for Kyverno policy recommendation

Vyom-Yadav:
Signed-off-by: Vyom-Yadav [jackhammervyom@gmail.com](mailto:jackhammervyom@gmail.com)

---

Kyverno policy recommendation is spread across 3 PRs including this one:

- https://github.com/kubearmor/policy-templates/pull/1037
- https://github.com/kubearmor/kubearmor-client/pull/303

---

### Explanation:

Kyverno policy recommendation dynamically recommends Kyverno policies based on container runtime data.

The dynamically recommended policy right now:

- Restrict automounting of service account token.

---

### How it works:

1. Policies are fetched from the latest release of https://github.com/kubearmor/policy-templates (the path is modified at present to my fork).
2. Algorithm determines whether the service account token is being used in pods in the deployment.
3. If it is unused, `restrict-automount-sa-token` is generated and inserted into the DB.
4. Since the policy recommendation is a cron job, if in subsequent runs, if container runtime data points towards usage of service account token, the policy for that particular deployment, if present, is deleted.

---

### Nuances:

1. **Policy present in `policy-templates` matches `Pod`s, but we generate a policy for `Deployment`s.**

[Kyverno autogen](https://kyverno.io/docs/writing-policies/autogen/) does not work when label selectors are used, see https://github.com/kyverno/kyverno/issues/4410.

Since we right now only recommend policies for Deployments (this logic will be changed when we shift to owner references), `autogen` for conversion from `Pod` -> `Deployment` is internally implemented. This can be easily extended for other owner references too if Kyverno does not support `autogen` in the future too.

2. **Kyverno policies are added to `policy_yaml` DB.**

As internally discussed, we will be adding policies to `policy_yaml` so we can extract policies from a single place. Other generated policies are right now added to both `policy_yaml` and other DB's like `system_policy`, `network_policy`, etc. The worker protobuf fetches system policies from `system_policy`, network policies from `network_policy`, **and admission controller policies from `policy_yaml` table. There is no `admission_controller_policy table`.**

3. **Worker has no control over admission controller policy recommendation.**

The [worker](https://github.com/accuknox/discovery-engine/blob/dev/src/protobuf/v1/worker/worker.proto) cannot stop or start the admission controller policy recommendation since it is a part of recommend cronjob. The recommend cronjob is automatically started when DE starts.

4. **The algorithm makes a few assumptions while checking whether SA token is mounted or not.** 

- Removing part of path returned by summary matching regular expression `..[0-9]{4}_[0-9]{2}_[0-9]{2}.*`: 

This is a known k8s feature/issue, read more about it at https://stackoverflow.com/q/50685385/15412365. Since KubeArmor resolved symlinks, this is required to match `VolumeMount` path.

- Removing the first element of the path if the complete path doesn't match:

Sometimes when SA account token is accessed at `/var/run/secrets/kubernetes.io/serviceaccount/token`, the path shown in summary is `/run/secrets/kubernetes.io/serviceaccount/..2023_03_28_12_23_09.412453730/token`. This is most likely a bug in symlink resolution, the correct path while resolving symlinks should be `/var/run/secrets/kubernetes.io/serviceaccount/..2023_03_28_12_23_09.412453730/token`. Read the discussion at https://accuknox.slack.com/archives/C0229TD83RA/p1680013050747849

---

Process of merging PRs:

1. Merge policy templates PR and publish a new release, so this PR can leverage those. The code won't break even if no Kyverno policies are present in a release.
2. Merge this PR.
3. Merge `kubearmor-client` PR.

---

Testing:

- Unit tests for validating policy recommendation algorithms have been added.


Vyom-Yadav:
Final recommended policy example:
```yaml
apiVersion: kyverno.io/v1
kind: Policy
metadata:
  annotations:
    policies.kyverno.io/minversion: 1.6.0
    recommended-policies.kubearmor.io/description: Don't mount service account token
      when it is not needed
    recommended-policies.kubearmor.io/description-detailed: If the Service Account
      Token is not used by a pod, then it should not be automounted. Service account
      token provide access to the kubeapi-server which potentially increases the surface
      area of attack.
    recommended-policies.kubearmor.io/tags: AUTOMOUNT_SERVICE_ACCOUNT
    recommended-policies.kubearmor.io/title: Restrict Auto-Mount of Service Account
      Tokens
  name: nginx-restrict-automount-sa-token
  namespace: default
spec:
  background: true
  rules:
  - match:
      any:
      - resources:
          kinds:
          - Deployment
    name: validate-automountServiceAccountToken
    preconditions:
      all:
      - key: '{{ request.operation || ''BACKGROUND'' }}'
        operator: NotEquals
        value: DELETE
      - key: '{{ request.object.spec.template.metadata.labels.app || '''' }}'
        operator: Equals
        value: nginx
      - key: '{{ request.object.spec.template.metadata.labels.foo || '''' }}'
        operator: Equals
        value: faa
    validate:
      message: Auto-mounting of Service Account tokens is not allowed.
      pattern:
        spec:
          template:
            spec:
              automountServiceAccountToken: "false"
  validationFailureAction: Audit

```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
nyrahul, line 1: Please fix the lints.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	if err := libs.UpdateOrInsertPolicyYamls(CfgDB, res); err != nil {
+		log.Error().Msgf(err.Error())
+	}
+}
+
+func GetAdmissionControllerPolicy(namespace, clusterName, labels string) []kyvernov1.Policy {
+	var kyvernoPolicies []kyvernov1.Policy
+	policyYamls, err := libs.GetPolicyYamls(CfgDB, types.PolicyTypeAdmissionController)
```
nyrahul: Why to fetch all the policy yamls ?? Can we pass the namespace, clusterName to this API and only fetch the relevant policies?

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+func ConvertPoliciesToWorkerResponse(policies []kyvernov1.Policy) *wpb.WorkerResponse {
+	var response wpb.WorkerResponse
+
+	for i := range policies {
+		kyvernoPolicy := wpb.Policy{}
+
+		val, err := json.Marshal(&policies[i])
+		if err != nil {
+			log.Error().Msgf("kyvernoPolicy json marshal failed err=%v", err.Error())
+		}
```
nyrahul:
```suggestion
		if err != nil {
			log.Error().Msgf("kyvernoPolicy json marshal failed err=%v", err.Error())
			continue
		}
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+				log.Warn().Msg(err.Error())
+				return true
+			}
+			sumResponses = append(sumResponses, sumResp)
+		}
+		return serviceAccountTokenUsed(containersSATokenMountPath, sumResponses)
+	} else {
+		log.Warn().Msg("No pods found for the given labels")
+	}
+
```
nyrahul:
```suggestion
	}
	log.Warn().Msg("No pods found for the given labels")

```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+			for _, projectedSources := range volume.Projected.Sources {
+				serviceAccountToken := projectedSources.ServiceAccountToken
+				if serviceAccountToken != nil {
+					tokenPath = serviceAccountToken.Path
+					projectedVolumeName = volume.Name
+					break
+				}
+			}
+		}
+	}
```
nyrahul:
```suggestion
	for _, volume := range volumes {
		if volume.Projected == nil {
		   continue
		}
		for _, projectedSources := range volume.Projected.Sources {
			serviceAccountToken := projectedSources.ServiceAccountToken
			if serviceAccountToken != nil {
				tokenPath = serviceAccountToken.Path
				projectedVolumeName = volume.Name
				break
			}
		}
		if tokenPath != "" {
		    break
		}
	}
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+
+	if tokenPath == "" || projectedVolumeName == "" {
+		return result,
+			fmt.Errorf("service account token not mounted for %s in namespace %s", pod.Name, pod.Namespace)
```
nyrahul, line 237: This is not an error and do not use fmt to print the error.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		return serviceAccountTokenUsed(containersSATokenMountPath, sumResponses)
+	} else {
+		log.Warn().Msg("No pods found for the given labels")
+	}
+
+	return true
+}
+
+func getSATokenMountPath(pod *corev1.Pod) ([]containerMountPathServiceAccountToken, error) {
```
nyrahul, line 213: What is the point of returning an array of `containerMountPathServiceAccountToken`? Based on the logic of the function there can atmost be one service account token associated with the pod.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+		volumeMounts := container.VolumeMounts
+		for _, volumeMount := range volumeMounts {
+			if volumeMount.Name == projectedVolumeName {
+				result = append(result, containerMountPathServiceAccountToken{
+					podName:          pod.Name,
+					podNamespace:     pod.Namespace,
+					containerName:    container.Name,
+					saTokenMountPath: volumeMount.MountPath + string(filepath.Separator) + tokenPath,
+				})
+			}
```
nyrahul, line 251: Assuming that we return only a single element in this function.
```suggestion
			if volumeMount.Name == projectedVolumeName {
				return containerMountPathServiceAccountToken{
					podName:          pod.Name,
					podNamespace:     pod.Namespace,
					containerName:    container.Name,
					saTokenMountPath: volumeMount.MountPath + string(filepath.Separator) + tokenPath,
				}
			}
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+	serviceAccountTokenUsed := false
+	for _, containerSATokenMountPath := range containersSATokenMountPath {
+		for _, sumResp := range sumResponses {
+			for _, fileData := range sumResp.FileData {
+				if sumResp.ContainerName == containerSATokenMountPath.containerName {
+					// Even if one container uses the service account token, we should allow auto mounting
+					if matchesSATokenPath(containerSATokenMountPath.saTokenMountPath, fileData.Destination) {
+						serviceAccountTokenUsed = true
+						break
+					}
```
nyrahul:
```suggestion
					if matchesSATokenPath(containerSATokenMountPath.saTokenMountPath, fileData.Destination) {
						return true
					}
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+				}
+			}
+		}
+	}
+	return serviceAccountTokenUsed
```
nyrahul:
```suggestion
	return false
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func matchesSATokenPath(saTokenPath, sumRespPath string) bool {
+	result := false
```
nyrahul:
```suggestion
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Typical SA token path /var/run/secrets/kubernetes.io/serviceaccount/token
+	// is also compared after removing first part i.e. /run/secrets/kubernetes.io/serviceaccount/token
+	// due to KubeArmor symlink resolution issue
+	saTokenPathWithoutFirstPathPart := strings.Join(append(saTokenPathParts[0:1],
+		saTokenPathParts[2:]...), string(filepath.Separator))
+
+	if saTokenPath == sumpRespPathWithoutDoubleDot ||
+		saTokenPathWithoutFirstPathPart == sumpRespPathWithoutDoubleDot {
+		result = true
```
nyrahul:
```suggestion
		return true
```

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	return result
```
nyrahul:
```suggestion
	return false
```

src/config/configManager.go
```diff
@@ -208,10 +210,11 @@ func LoadConfigFromFile() {
...
 	CurrentCfg.ConfigRecommendPolicy = types.ConfigRecommendPolicy{
-		CronJobTimeInterval:     "@every " + viper.GetString("recommend.cron-job-time-interval"),
-		OneTimeJobTimeSelection: "", // e.g., 2021-01-20 07:00:23|2021-01-20 07:00:25
-		OperationMode:           viper.GetInt("recommend.operation-mode"),
-		RecommendHostPolicy:     viper.GetBool("recommend.recommend-host-policy"),
+		CronJobTimeInterval:                "@every " + viper.GetString("recommend.cron-job-time-interval"),
+		OneTimeJobTimeSelection:            "", // e.g., 2021-01-20 07:00:23|2021-01-20 07:00:25
+		OperationMode:                      viper.GetInt("recommend.operation-mode"),
+		RecommendHostPolicy:                viper.GetBool("recommend.recommend-host-policy"),
+		RecommendAdmissionControllerPolicy: viper.GetBool("recommend.recommend-admission-controller-policy"),
```
nyrahul:
```suggestion
		RecommendAdmissionControllerPolicy: viper.GetBool("recommend.admission-controller-policy"),
```
can we us simplified config name?

src/libs/common.go
```diff
@@ -151,7 +151,7 @@ func SetDefaultConfig() {
 	viper.SetDefault("recommend.cron-job-time-interval", "1h0m00s")
 	viper.SetDefault("recommend.operation-mode", 1)
 	viper.SetDefault("recommend.recommend-host-policy", true)
-
+	viper.SetDefault("recommend.recommend-admission-controller-policy", true)
```
nyrahul:
```suggestion
	viper.SetDefault("recommend.admission-controller-policy", true)
```

src/recommendpolicy/downloadTemplates.go
```diff
@@ -18,9 +19,9 @@ import (
 )
 
 const (
-	org   = "kubearmor"
+	org   = "Vyom-Yadav"
```
nyrahul: Use appropriate repo name here.

src/recommendpolicy/helperFunctions.go
```diff
@@ -96,6 +100,69 @@ func generatePolicy(name, namespace string, labels LabelMap) ([]types.KnoxSystem
...
+			case "restrict-automount-sa-token":
+				if !admissioncontrollerpolicy.ShouldSATokenBeAutoMounted(namespace, labels) {
+					log.Info().Msgf("Generating admission controller policy for deployment: %v in namespace: %v", name, namespace)
+					policies = append(policies, createRestrictAutomountSATokenPolicy(ms, name, namespace, labels))
+				} else {
+					policiesToBeDeleted = append(policiesToBeDeleted, name+"-"+ms.Name)
+				}
+			}
+		}
+	}
```
nyrahul:
```suggestion
	for ; err == nil; ms, err = getNextRule(&idx) {
		if ms.KyvernoPolicySpec == nil {
		    continue
		}
		switch ms.Name {
		case "restrict-automount-sa-token":
			if !admissioncontrollerpolicy.ShouldSATokenBeAutoMounted(namespace, labels) {
				log.Info().Msgf("Generating admission controller policy for deployment: %v in namespace: %v", name, namespace)
				policies = append(policies, createRestrictAutomountSATokenPolicy(ms, name, namespace, labels))
			} else {
				policiesToBeDeleted = append(policiesToBeDeleted, name+"-"+ms.Name)
			}
		}
	}
```
reduce nesting

src/recommendpolicy/helperFunctions.go
```diff
@@ -96,6 +100,69 @@ func generatePolicy(name, namespace string, labels LabelMap) ([]types.KnoxSystem
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name + "-" + ms.Name,
+			Annotations: map[string]string{
+				"recommended-policies.kubearmor.io/title":                "Restrict Auto-Mount of Service Account Tokens",
+				"policies.kyverno.io/minversion":                         "1.6.0",
+				"recommended-policies.kubearmor.io/description-detailed": ms.Description.Detailed,
+				"recommended-policies.kubearmor.io/tags":                 strings.Join(ms.KyvernoPolicyTags, ","),
+				"recommended-policies.kubearmor.io/description":          ms.Description.Tldr,
+			},
```
nyrahul, line 143: I was hoping all this was read from the kyverno policy in the policy-template.

src/recommendpolicy/recommendPolicy.go
```diff
@@ -136,31 +136,63 @@ func RecommendPolicyMain() {
...
+			} else if len(nsNotFilterAdmissionControllerPolicy) > 0 {
+				for _, ns := range nsNotFilterAdmissionControllerPolicy {
+					if d.Namespace != ns {
+						generateAdmissionControllerPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
+					}
+				}
+			} else {
+				generateAdmissionControllerPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
+			}
+		}
```
nyrahul: Please refactor this code on following lines...
```suggestion
	    if recommendAdmissionControllerPolicy && !isNsFiltered(d.Namespac) {
			generateAdmissionControllerPolicy(d.Name, d.Namespace, d.Spec.Template.Labels)
		}
```
implement `isNsFiltered()` to check if the namespace is filtered.

nyrahul (changes requested):

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: > This is not an error

The method `getSATokenMountPath(..)` gets the mount path of the SA token, if it is not mounted, it returns an error. 

> fmt to print the error.

The error is not printed to the console with `fmt`, it is just formatted with `fmt`.

src/recommendpolicy/helperFunctions.go
Vyom-Yadav: It can be. 

But the general pattern that we follow for policies is that we store just the spec of the policy after downloading them. 
```go
type MatchSpec struct {
	Name              string          `json:"name" yaml:"name"`
	Precondition      []string        `json:"precondition" yaml:"precondition"`
	Description       Description     `json:"description" yaml:"description"`
	Yaml              string          `json:"yaml" yaml:"yaml"`
	Spec              KnoxSystemSpec  `json:"spec,omitempty" yaml:"spec,omitempty"`
	Kind              string          `json:"kind,omitempty" yaml:"kind,omitempty" bson:"kind,omitempty"`
	KyvernoPolicySpec *kyvernov1.Spec `json:"kyvernoPolicySpec,omitempty" yaml:"kyvernoPolicySpec,omitempty"`
	KyvernoPolicyTags []string        `json:"kyvernoPolicyTags,omitempty" yaml:"kyvernoPolicyTags,omitempty"`
}
```

One alternative can be storing the entire policy instead of the spec, other one could be introducing a new field and ignore it with a temporary struct while marshaling.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: My bad, fixed.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: There can be at most one SA token, but for pod with multiple containers, SA token is mounted on both. Yes, the path is the same, but this model gives more flexibility and returns the mount point of SA token per container and not per pod.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

src/config/configManager.go
Vyom-Yadav: Done, changed both `host-policy` and `admission-controller-policy`

src/libs/common.go
Vyom-Yadav: Done.

src/recommendpolicy/downloadTemplates.go
Vyom-Yadav: This is just for testing if you want to `gh checkout...` or test these changes locally as mentioned in the PR description. 

Changed this too.

src/recommendpolicy/helperFunctions.go
Vyom-Yadav: Done.

src/recommendpolicy/recommendPolicy.go
Vyom-Yadav: Done.

src/admissioncontrollerpolicy/admissionControllerPolicy.go
Vyom-Yadav: Done.

nyrahul (approved):
Awesome PR. LGTM. 👍

src/admissioncontrollerpolicy/admissionControllerPolicy.go
```diff
@@ -0,0 +1,309 @@
...
+		},
+	}
+	return newPattern
+}
+
+// ShouldSATokenBeAutoMounted returns true if service account token should be auto mounted
+func ShouldSATokenBeAutoMounted(namespace string, labels types.LabelMap) bool {
+	client := cluster.ConnectK8sClient()
+
+	podList, err := client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
```
vishnusomank, line 173: can we have this handled in the [k8sClientHandler.go](https://github.com/accuknox/discovery-engine/blob/dev/src/cluster/k8sClientHandler.go) file?

Vyom-Yadav: Yes, the function can be transferred to that file, but it won't make much sense as this is something particular to admission controller policies. We just connect to the client to list pods.

vishnusomank (approved):
LGTM!

nyrahul (approved):

PR #702, opened 2023-04-03
feat(license): add license feature

rajaSahil:
- Add License feature
- Add protobuf for licensing


src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,101 @@ func GetKubearmorRelayURL() string {
...
+	kubeSystem, err := k8sClient.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	uuid := string(kubeSystem.GetUID())
+
+	return uuid, nil
+
+}
```
nyrahul:
```suggestion
	return string(kubeSystem.GetUID()), nil
}
```

src/license/license.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/mervick/aes-everywhere/go/aes256"
+	"github.com/rs/zerolog/log"
+	"k8s.io/client-go/kubernetes"
+	"strings"
+	"time"
+)
+
+// For testing purpose
+var publicKey = "-----BEGIN PUBLIC KEY-----\nMIGeMA0GCSqGSIb3DQEBAQUAA4GMADCBiAKBgHUc95xoPHqsuC3zLfCSHHJ9F/Gx\nlJdyBkns1wDYCLY8yX1vvZndfDP9br3dbFKOaYOYmF9e0gKcDpGItdBQe+TVX9ol\nM3S23yD/xHNKw+f88KjI0dPnj3IRgqajd5eBMhNNugRFzRKWBBLCflukm7CfjzUP\nX1jQ/NCkoTwjScpJAgMBAAE=\n-----END PUBLIC KEY-----"
```
rajaSahil, line 17: This is for testing purpose

nyrahul, line 17: I am not sure how is this public key used?

src/license/license.go
```diff
@@ -0,0 +1,254 @@
...
+
+func (t *Token) getFeatures() ([]string, error) {
+	features := t.claims.Features
+	return features, nil
+}
+
+// WatchFeatures to validate license and watch features
+func (cfg *LicenseConfig) WatchFeatures() bool {
+
+	for {
```
nyrahul, line 230: If the license check fails, this would go in tight loop!

nyrahul (changes requested):

src/license/license.go
rajaSahil: This is being used to verify signature for jwt token.

src/license/license.go
rajaSahil: Have added 5 sec sleep.

src/license/license.go
```diff
@@ -0,0 +1,295 @@
...
+		k8sClient: k8sClient,
+		Tkn:       nil,
+		Lcs:       nil,
+	}
+}
+
+// CheckLicenseSecret to fetch license secret and validate
+func CheckLicenseSecret() error {
+	log.Info().Msgf("fetching license secrets to validate discovery-engine licensing")
+	secret, err := cluster.GetSecrets(LCfg.k8sClient, "app=discovery-engine")
```
vishnusomank: @rajaSahil  should we hardcode the labels for the discovery engine here? shouldn't it be configurable using the configmap?

src/license/license.go
```diff
@@ -0,0 +1,295 @@
...
+		log.Error().Msgf("error while parsing jwt token, error: %s", err.Error())
+		return nil, err
+	}
+
+	Tkn := &Token{
+		jwt:    jwtToken,
+		claims: claims,
+	}
+
+	err = Tkn.validateClaims(userId)
```
vishnusomank: @rajaSahil why are we calling `Tkn.validateClaims(userId)` can't we directly call `Tkn.validateUserId(userId)` ?

src/license/license.go
```diff
@@ -0,0 +1,295 @@
...
+			return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+		}
+		return key, nil
+	})
+	if err != nil {
+		log.Error().Msgf("error while parsing jwt token, error: %s", err.Error())
+		return nil, err
+	}
+
+	Tkn := &Token{
```
vishnusomank: since we are not exporting this variable why don't we use a smaller case?

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,106 @@ func GetKubearmorRelayURL() string {
...
+
+// CreateLicenseSecret to create license secret
+func CreateLicenseSecret(k8sClient *kubernetes.Clientset, key string, userId string) (*v1.Secret, error) {
+	if k8sClient == nil {
+		return nil, errors.New("k8s client not created")
+	}
+
+	secret, err := GetSecrets(k8sClient, "app=discovery-engine")
```
vishnusomank: @rajaSahil should we hardcode the labels for the discovery engine here? shouldn't it be configurable using the configmap?

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,106 @@ func GetKubearmorRelayURL() string {
...
+	if secret != nil {
+		log.Info().Msgf("secrets already exists for discovery-engine license for user-id: %s", userId)
+		return secret, nil
+	}
+	t := true
+
+	secretSpec := v1.Secret{
+
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "discovery-engine",
```
vishnusomank: can we have this name as a constant and standardize it to something like `discovery-engine-license-key` or similar?

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,106 @@ func GetKubearmorRelayURL() string {
...
 	return url
 }
+
+// GetSecrets to get secrets
+func GetSecrets(k8sClient *kubernetes.Clientset, label string) (*v1.Secret, error) {
+	if k8sClient == nil {
+		return nil, errors.New("k8s client not created")
+	}
+
+	secrets, err := k8sClient.CoreV1().Secrets("").List(context.Background(), metav1.ListOptions{
```
vishnusomank: Since we are creating this secret under 'kube-system' namespace with a predefined name can't we use `Secrets.Get` to get the secret?

vishnusomank (changes requested):

src/license/license.go
rajaSahil: I don't think configmap is required here, since we only need a label and a name.
We can make it constant and use it from the code itself.

src/license/license.go
rajaSahil: No, right now I have only validated userID but there are going to be other validation also.

src/cluster/k8sClientHandler.go
rajaSahil: Same comment https://github.com/accuknox/discovery-engine/pull/702#discussion_r1163845462

src/cluster/k8sClientHandler.go
rajaSahil: Yes, I have made this a constant, will be pushing the commit.

src/cluster/k8sClientHandler.go
rajaSahil: Yes, We can.
But I think using label would help us to avoid duplicacy.

vishnusomank (approved):

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,98 @@ func GetKubearmorRelayURL() string {
...
 	return url
 }
+
+// GetSecrets to get secrets
+func GetSecrets(k8sClient *kubernetes.Clientset, label string, name string) (*v1.Secret, error) {
+	if k8sClient == nil {
+		return nil, errors.New("k8s client not created")
+	}
+
+	secrets, err := k8sClient.CoreV1().Secrets("kube-system").Get(context.Background(), name, metav1.GetOptions{})
```
nyrahul: We should not create any resources in the kube-system namespace. At best, we should keep this configurable.

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,98 @@ func GetKubearmorRelayURL() string {
...
+
+	if err != nil {
+		log.Error().Msgf("error while getting license secrets with name: %s from kube-system namespace, error: %s", name, err.Error())
+		return nil, err
+	}
+	if secrets == nil {
+		return nil, nil
+	}
+
+	return secrets, nil
```
nyrahul:
```suggestion
	return secrets, nil
```
`secrets` will be `nil` in case of failure anyways.

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,98 @@ func GetKubearmorRelayURL() string {
...
+	return secret, nil
+}
+
+// DeleteSecrets to delete secrets
+func DeleteSecrets(k8sClient *kubernetes.Clientset, name string) error {
+	if k8sClient == nil {
+		return errors.New("k8s client not created")
+	}
+
+	err := k8sClient.CoreV1().Secrets("kube-system").Delete(context.Background(), name, metav1.DeleteOptions{})
```
nyrahul: Remove hard-coded `kube-system` namespace.

src/cluster/k8sClientHandler.go
```diff
@@ -498,3 +502,98 @@ func GetKubearmorRelayURL() string {
...
+	return nil
+}
+
+// GetKubeSystemUUID get kube-system namespace uuid
+func GetKubeSystemUUID(k8sClient *kubernetes.Clientset) (string, error) {
+	if k8sClient == nil {
+		return "", errors.New("k8s client not created")
+	}
+
+	kubeSystem, err := k8sClient.CoreV1().Namespaces().Get(context.Background(), "kube-system", metav1.GetOptions{})
```
nyrahul, line 590: remove hardcoded `kube-system` namespace.

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+	"strings"
+	"time"
+)
+
+// For testing purpose
+var publicKey = "-----BEGIN PUBLIC KEY-----\nMIGeMA0GCSqGSIb3DQEBAQUAA4GMADCBiAKBgHUc95xoPHqsuC3zLfCSHHJ9F/Gx\nlJdyBkns1wDYCLY8yX1vvZndfDP9br3dbFKOaYOYmF9e0gKcDpGItdBQe+TVX9ol\nM3S23yD/xHNKw+f88KjI0dPnj3IRgqajd5eBMhNNugRFzRKWBBLCflukm7CfjzUP\nX1jQ/NCkoTwjScpJAgMBAAE=\n-----END PUBLIC KEY-----"
+
+const (
+	discoveryEngineSecretName = "discovery-engine-license"
+	licenseSecLabel           = "app=discover-engine"
```
nyrahul:
```suggestion
	licenseSecLabel           = "app=discovery-engine"
```

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	err = l.ValidateLicense()
+	LCfg.Lcs = l
+
+	if err != nil {
+		log.Error().Msgf("error while validating license retrieved through secrets, error: %s", err.Error())
+		return err
+	}
+	// Initialize to global config only after validation is done.
+	LCfg.Lcs = l
```
nyrahul: This is already handled on line 66.
```suggestion
```

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+	if secret == nil {
+		return errors.New("license secret doesn't exist for discovery-engine")
+	}
+
+	l := &License{
+		UserId: string(secret.Data["user-id"]),
+		Key:    string(secret.Data["key"]),
+	}
+	err = l.ValidateLicense()
+	LCfg.Lcs = l
```
nyrahul, line 68: This assignment is already done in `ValidateLicense()`.
```suggestion
```

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+	return t, err
+}
+
+func (t *Token) validateClaims(userId string) error {
+	err := t.validateUserId(userId)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
nyrahul:
```suggestion
func (t *Token) validateClaims(userId string) error {
	return t.validateUserId(userId)
}
```

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	return nil
+
+}
+
+func (t *Token) getFeatures() ([]string, error) {
+	features := t.claims.Features
+	return features, nil
+}
+
```
nyrahul:
```suggestion
func (t *Token) getFeatures() ([]string, error) {
	return t.claims.Features, nil
}
```

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	return false
+}
+
+func (cfg *LicenseConfig) checkExistingLicense() bool {
+	if cfg.Lcs != nil || cfg.Tkn != nil {
+		return true
+	}
+	return false
+}
```
nyrahul:
```suggestion
func (cfg *LicenseConfig) checkExistingLicense() bool {
	return (cfg.Lcs != nil || cfg.Tkn != nil)
}
```

src/license/license.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (cfg *LicenseConfig) removeLicenseSecretsConfig() error {
+	err := cluster.DeleteSecrets(LCfg.k8sClient, discoveryEngineSecretName)
+	if err != nil {
+		return err
+	}
+	cfg.Lcs = nil
+	cfg.Tkn = nil
+	return nil
+}
```
nyrahul:
```suggestion
func (cfg *LicenseConfig) removeLicenseSecretsConfig() error {
	cfg.Lcs = nil
	cfg.Tkn = nil
	return cluster.DeleteSecrets(LCfg.k8sClient, discoveryEngineSecretName)
}
```

src/main.go
```diff
@@ -40,23 +43,53 @@ func init() {
...
+	server = grpcserver.AddLicenseServer(server)
+
+	// check for license secret, if exist then validate
+	err := license.CheckLicenseSecret()
+
+	if err != nil {
+		log.Error().Msgf("error while validating license secrets for discovery engine, error: %s", err.Error())
+		go serve(lis, server)
+		_ = license.LCfg.WatchFeatures()
+		os.Exit(1)
```
nyrahul, line 68: I  don't understand the point of spawning a go routine when that go routine would not be handled since the code is going to do os.Exit().

src/server/grpcServer_test.go
```diff
@@ -7,6 +7,8 @@ import (
 )
 
 func TestGetNewServer(t *testing.T) {
-	server := GetNewServer()
+	server := StartGrpcServer()
+	server = AddLicenseServer(server)
+	server = AddServers(server)
 	assert.NotNil(t, server)
```
nyrahul:
```suggestion
	assert.NotNil(t, AddServers(AddLicenseServer(StartGrpcServer())))
```

nyrahul (changes requested):

src/cluster/k8sClientHandler.go
rajaSahil: Sure, by default where should we create the secrets @nyrahul ?

src/license/license.go
rajaSahil: Assignment will only happen if there are no errors in ValidateLicense, but if there comes any error assignment won't happen.

src/main.go
rajaSahil: grpc server starts on the new go routine and as soon as a license is installed we exit from the code.

src/cluster/k8sClientHandler.go
Ankurk99: accuknox-agents can be kept as the default ns

src/license/license.go
nyrahul: But the assignment was already done inside `ValidateLicense()`

nyrahul (changes requested):
There are few comments from my past reviews that are not handled. Or am i mistaken?

src/cluster/k8sClientHandler.go
```diff
@@ -499,3 +503,95 @@ func GetKubearmorRelayURL() string {
...
+// GetSecrets to get secrets
+func GetSecrets(k8sClient *kubernetes.Clientset, label string, namespace string, name string) (*v1.Secret, error) {
+	if k8sClient == nil {
+		return nil, errors.New("k8s client not created")
+	}
+
+	secrets, err := k8sClient.CoreV1().Secrets(namespace).Get(context.Background(), name, metav1.GetOptions{})
+
+	if err != nil {
+		log.Error().Msgf("error while getting license secrets with name: %s from kube-system namespace, error: %s", name, err.Error())
```
nyrahul:
```suggestion
		log.Error().Msgf("error while getting license secrets with name: %s from %s namespace, error: %s", name, namespace, err.Error())
```

nyrahul (changes requested):

rajaSahil:
https://github.com/accuknox/discovery-engine/pull/702#discussion_r1165791805 But assignment only happens if ValidateLicense() doesn't throw any error.

nyrahul (approved):

PR #707, opened 2023-04-11
New Github Workflow for agents

arundevops07:
ci-go.yaml,release.yaml and others got deleted In order to avoid redundant workflows, since go -checks added in action.yaml itself.


Ankurk99:
@arunaccuknox Can we close this since we have those files restored?

PR #712, opened 2023-04-18
feat(license): fix code

rajaSahil:
- Fix code

### DIFF
--- a/src/cluster/k8sClientHandler.go
+++ b/src/cluster/k8sClientHandler.go
@@ -512,9 +512,13 @@ func GetSecrets(k8sClient *kubernetes.Clientset, label string, namespace string,
 
 	secrets, err := k8sClient.CoreV1().Secrets(namespace).Get(context.Background(), name, metav1.GetOptions{})
 
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "not found") {
 		log.Error().Msgf("error while getting license secrets with name: %s from %s namespace, error: %s", name, namespace, err.Error())
 		return nil, err
+	}
+
+	if secrets != nil && secrets.GetName() == "" {
+		return nil, nil
 	}
 
 	return secrets, nil


src/cluster/k8sClientHandler.go
```diff
@@ -512,11 +512,15 @@ func GetSecrets(k8sClient *kubernetes.Clientset, label string, namespace string,
 
 	secrets, err := k8sClient.CoreV1().Secrets(namespace).Get(context.Background(), name, metav1.GetOptions{})
 
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "not found") {
```
Ankurk99, line 515: What are we trying to do here? What's the case when the error says "not found" and should we be handling that case separately?

nyrahul: Yeah this is not a good practice.

nyrahul (changes requested):

src/cluster/k8sClientHandler.go
```diff
@@ -512,11 +512,15 @@ func GetSecrets(k8sClient *kubernetes.Clientset, label string, namespace string,
...
 		log.Error().Msgf("error while getting license secrets with name: %s from %s namespace, error: %s", name, namespace, err.Error())
 		return nil, err
 	}
 
+	if secrets.GetName() == "" {
```
vishnusomank: If the `secrets` variable is `nil` this will cause a runtime error, right?

src/cluster/k8sClientHandler.go
rajaSahil: There are errors such as `secrets "discov │
│ ery-engine-license" is forbidden: User "system:serviceaccount:accuknox-agents:discovery-engine" cannot get resource "secrets" in API group "" in the namespace "accuknox-agents"` which comes because of rbac issues.

rajaSahil: To isolate these errors, I have added this check.

vishnusomank: if there is an error that means we don't have data in secret, right? And since we are only printing the error we shouldn't be bothered with the different cases of error that might get generated. Or are we going to handle different types of errors?

rajaSahil: There might be case, when secrets got created but after some restarts or change in cluster rbac policies we started receiving this error. In this case, if we don't handle this error, then new license validation will happen and it will fail at the last time.

vishnusomank (approved):

src/cluster/k8sClientHandler.go
Ankurk99: I see the reasoning here. We need to fix the rbac rules for discovery-engine then. But if you want to handle that separately then that's also fine ig.
@nyrahul Wdyt?

rajaSahil: We can use the same serviceaccount that we use for accuknox-agents operator. @Ankurk99

Ankurk99: Okay, we can fix that in a separate PR. Otherwise LGTM

Ankurk99 (approved):

nyrahul (approved):

PR #713, opened 2023-04-21
fix(summary): exclude `container_id` from `system_summary` table

### LINKED ISSUE #668: [bug] multiple entries for same process in karmor summary (vault app)
Screenshot:
image

To Reproduce this issue:
```console
$ helm repo add hashicorp https://helm.releases.hashicorp.com
"hashicorp" has been added to your repositories

$ helm install vault hashicorp/vault
```
and run `karmor summary`

yasin-cs-ko-ak:
## Issue:
Repeated entries in `system_summary `table due to different `container_id` when pod restarts.

## Description:
Every time a pod restarts, it creates a new `container_id.` This results in repeated entries in the `system_summary` table for the same pod and container, causing issues with data analysis when we are using karmor summary in CLI.

## Fix:
To prevent repeated entries in the `system_summary` table, this PR exclude the `container_id` field from the `upsertSysSummarySQL` function. Instead, the function will now use the `podname` field to check for existing entries and update them accordingly. This will ensure that only one entry exists in the table for each unique pod, regardless of how many times the container is restarted.

### DIFF
--- a/src/libs/dbHandler.go
+++ b/src/libs/dbHandler.go
@@ -3,6 +3,8 @@ package libs
 import (
 	"database/sql"
 	"errors"
+	"sort"
+	"strings"
 
 	cfg "github.com/accuknox/auto-policy-discovery/src/config"
 	logger "github.com/accuknox/auto-policy-discovery/src/logging"
@@ -425,8 +427,13 @@ func UpsertSystemSummary(cfg types.ConfigDB, summaryMap map[types.SystemSummary]
 }
 
 func upsertSysSummarySQL(db *sql.DB, summary types.SystemSummary, timeCount types.SysSummaryTimeCount) error {
+	// sorts pod labels upon pod restart
+	sortedLabels := strings.Split(summary.Labels, ",")
+	sort.Strings(sortedLabels)
+	summary.Labels = strings.Join(sortedLabels, ",")
+
 	queryString := `cluster_name = ? and cluster_id = ? and workspace_id = ? and namespace_name = ? and namespace_id = ? and container_name = ? and container_image = ? 
-					and container_id = ? and podname = ? and operation = ? and labels = ? and deployment_name = ? and source = ? and destination = ? 
+					and podname = ? and operation = ? and labels = ? and deployment_name = ? and source = ? and destination = ? 
 					and destination_namespace = ? and destination_labels = ? and type = ? and ip = ? and port = ? and protocol = ? and action = ? and bindport = ? and bindaddr = ?`
 
 	query := "UPDATE " + TableSystemSummarySQLite + " SET count=count+?, updated_time=? WHERE " + queryString + " "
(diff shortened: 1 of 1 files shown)


seswarrajan (changes requested):
Also remove ContainerID from types/observability.go Line: 105

yasin-cs-ko-ak:
> Also remove ContainerID from types/observability.go Line: 105

if we remove the type from there, container_id will be completely removed from the db after that we can't test anything. we are only excluding the container_id.

src/libs/dbHandler.go
```diff
@@ -425,8 +427,13 @@ func UpsertSystemSummary(cfg types.ConfigDB, summaryMap map[types.SystemSummary]
 }
 
 func upsertSysSummarySQL(db *sql.DB, summary types.SystemSummary, timeCount types.SysSummaryTimeCount) error {
+	// sorts pod labels upon pod restart
+	sortLabels := summary.Labels
+	sortedLabels := strings.Split(sortLabels, ",")
+	sort.Strings(sortedLabels)
```
vishnusomank, line 432: @yasin-cs-ko-ak  What is this code doing?

yasin-cs-ko-ak:
Before sorting the labels

before-labels

After sorting the labels

after-labels

cc: @vishnusomank

yasin-cs-ko-ak:
**Before this fix:**

before-fix

**After this fix:** (there is no repeated entries)

after-fix

cc: @vishnusomank

vishnusomank (approved):

nyrahul (approved):

seswarrajan (approved):
LGTM

seswarrajan:
The reason for removing container_id from system_summary table is that to avoid repetitive logs. Do we not need the same for observability as well. We can remove container_id in summary from being sent to DEO.

@nyrahul @yasin-cs-ko-ak @vishnusomank @wazir-ahmed @AmanKumar2696  WDYT ?

nyrahul:
What is the purpose of observability table vs system summary table? @seswarrajan

seswarrajan:
> What is the purpose of observability table vs system summary table? @seswarrajan

We do not have separate observability table @nyrahul 
What i meant to mention is that can we disable ContainerID field from being sent to DEO ?

AmanKumar2696:
> The reason for removing container_id from system_summary table is that to avoid repetitive logs. Do we not need the same for observability as well. We can remove container_id in summary from being sent to DEO.
> 
> @nyrahul @yasin-cs-ko-ak @vishnusomank @wazir-ahmed @AmanKumar2696 WDYT ?

In App Behaviour, we're not using ```container_id``` for any searching/filtering. Removing it should not affect any existing APIs.

yasin-cs-ko-ak:
We are only removing `container_id` from update query. So we can resolve this issue with minimal code changes.

PR #717, opened 2023-04-27
fix(license): minor code fix

rajaSahil:
- Minor code fix for license

### DIFF
--- a/src/license/license.go
+++ b/src/license/license.go
@@ -59,19 +59,34 @@ func CheckLicenseSecret() error {
 		return errors.New("license secret doesn't exist for discovery-engine")
 	}
 
-	l := &License{
+	if string(secret.Data["user-id"]) == "" || string(secret.Data["key"]) == "" {
+		err := fmt.Errorf("invalid secret exists for license")
+		log.Error().Msgf("error: %s", err)
+		return err
+	}
+
+	LCfg.Lcs = &License{
 		UserId: string(secret.Data["user-id"]),
 		Key:    string(secret.Data["key"]),
 	}
-	err = l.ValidateLicense()
-	// Initialize to global config only after validation is done.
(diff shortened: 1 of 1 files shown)


nyrahul (approved):
LGTM

src/license/license.go
```diff
@@ -75,6 +76,15 @@ func CheckLicenseSecret() error {
 	return nil
 }
 
+func (t *Token) checkTokenExpiration() error {
+	if t.checkExpiration() {
+		err := fmt.Errorf("license is expired, valid license doesn't exist with user-id: %s, key: %s and platform uuid: %s", LCfg.Lcs.UserId, LCfg.Lcs.Key, LCfg.Lcs.PlatformUUID)
```
seswarrajan, line 86: Do we need to print key in the error ?
We need only user and uuid to generate new token. I think only those 2 are required.

seswarrajan (changes requested):

src/license/license.go
rajaSahil: Printed it so user can validate license key that they used while installing.

src/license/license.go
```diff
@@ -59,13 +59,14 @@ func CheckLicenseSecret() error {
 		return errors.New("license secret doesn't exist for discovery-engine")
 	}
 
-	l := &License{
+	LCfg.Lcs = &License{
 		UserId: string(secret.Data["user-id"]),
```
vishnusomank, line 69: What if the secret exists and the required keys `user-id` or `key` is not found.? In this scenario, we can directly return an error and exit right?

rajaSahil: Ah, I didn't consider this edge case, thanks for pointing it out!

seswarrajan (approved):
LGTM

PR #721, opened 2023-05-04
feat(license): add config for license

rajaSahil:
- Add config to enable/disable license
- Add validate option

### DIFF
--- a/src/libs/common.go
+++ b/src/libs/common.go
@@ -152,6 +152,7 @@ func SetDefaultConfig() {
 	viper.SetDefault("recommend.operation-mode", 1)
 	viper.SetDefault("recommend.host-policy", true)
 	viper.SetDefault("recommend.admission-controller-policy", true)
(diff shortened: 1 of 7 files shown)


src/libs/common.go
```diff
@@ -222,7 +223,7 @@ func CheckCommandLineConfig() {
 		os.Exit(0)
 	}
 
-	viper.SetConfigName(GetEnv("CONF_FILE_NAME", "conf"))
+	viper.SetConfigName(GetEnv("CONF_FILE_NAME", "local"))
```
nyrahul: Lets not change this if it is not impacting ... there might be some auto-test frameworks that might break.

nyrahul (changes requested):
@rajaSahil , can you please add making the platform-uuid check optional as part of this PR?

nyrahul (approved):
LGTM

PR #722, opened 2023-05-09
fix(license): add license enable check

rajaSahil:
- Add license enable check


Ankurk99 (commented):
Change looks fine. Can you please also add the license enable config to `src/conf/local-file.yaml`? That's the one we use for running DE locally.

rajaSahil:
@Ankurk99 Addressed.

Ankurk99 (approved):
LGTM!

PR #723, opened 2023-05-10
Use deployment name value from kubearmor protobuf logs

seswarrajan:

### DIFF
--- a/deployments/helm/configmapfiles/discovery-engine/conf.yaml
+++ b/deployments/helm/configmapfiles/discovery-engine/conf.yaml
@@ -84,3 +84,6 @@ recommend:
 license:
   enabled: false
   validate: "user-id"
+
+# pprof
+pprof: false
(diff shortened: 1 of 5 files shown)


src/observability/summarizer.go
```diff
@@ -124,13 +122,8 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Alert) {
...
 		sysSummary.Severity = syslog.Severity
 		sysSummary.PolicyName = syslog.PolicyName
-		sysSummary.Deployment = ""
-
-		for _, d := range deployments {
-			if strings.Contains(syslog.Labels, d.Labels) && d.Namespace == syslog.NamespaceName {
-				sysSummary.Deployment = d.Name
-				break
-			}
+		if syslog.Owner.Ref == types.PodTypeDeployment {
```
vishnusomank: this will make the `sysSummary.Deployment` value `null` if the type is not deployment right? Shouldn't we have a parent name attached to each pod? even if it's not deployed using a deployment?

PR #735, opened 2023-05-28
Update Discovery Engine to get workload type from KubeArmor

AmanKumar2696:
- Parse updated kubearmor event structure to fetch workload type.

Signed off by: Aman Kumar <amankumar2696@gmail.com>


src/types/observability.go
```diff
@@ -94,39 +94,45 @@ type NetworkSummary struct {
...
-	PolicyName     string `json:"PolicyName,omitempty"`
+	ClusterName    string   `json:"ClusterName,omitempty"`
+	ClusterId      int32    `json:"ClusterId,omitempty"`
+	NamespaceName  string   `json:"Namespace,omitempty"`
+	NamespaceId    int32    `json:"NamespaceId,omitempty"`
+	ContainerName  string   `json:"ContainerName,omitempty"`
+	ContainerImage string   `json:"ContainerImage,omitempty"`
+	ContainerID    string   `json:"ContainerID,omitempty"`
+	Owner          PodOwner `json:"Owner,omitempty"`
+	PodName        string   `json:"PodName,omitempty"`
```
seswarrajan, line 110: @AmanKumar2696 I suggest do not add any new entries into the struct in between. Please add it last. 
Thanks.

src/protobuf/v1/publisher/publisher.proto
```diff
@@ -25,6 +31,7 @@ message SummaryResponse{
     string ContainerName = 5;
     string ContainerImage = 6;
     string ContainerId = 7;
+    Podowner Owner=32;
```
seswarrajan: @AmanKumar2696 The order is valid, but please add it to the last.

src/observability/summarizer.go
```diff
@@ -125,6 +125,9 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Alert) {
 		sysSummary.Severity = syslog.Severity
 		sysSummary.PolicyName = syslog.PolicyName
 		sysSummary.Deployment = ""
+		sysSummary.Owner.Ref = syslog.Owner.Ref
+		sysSummary.Owner.Name = syslog.Owner.Name
+		sysSummary.Owner.Namespace = syslog.Owner.Namespace
 
 		for _, d := range deployments {
```
seswarrajan: @AmanKumar2696 If we are receiving all details from owner reference struct. Then kindly remove the client-go calls to get the deployment data ( i.e code from line 132 to 137)

seswarrajan (changes requested):

src/types/observability.go
AmanKumar2696: Sure.

src/types/observability.go
```diff
@@ -94,39 +94,45 @@ type NetworkSummary struct {
 	UpdatedTime          int64  `json:"UpdatedTime,omitempty"`
 	Count                int32  `json:"Count,omitempty"`
 }
+type PodOwner struct {
+	Ref       string `json:"Ref,omitempty"`
+	Name      string `json:"Name,omitempty"`
+	Namespace string `json:"Namespace,omitempty"`
+}
```
wazir-ahmed:
```suggestion
type Workload struct {
	Type       string `json:"Type,omitempty"`
	Name      string `json:"Name,omitempty"`
}
```
- We don't need `Namespace` here since we already have `SystemSummary.NamespaceName`.

src/protobuf/v1/publisher/publisher.proto
```diff
@@ -17,6 +17,12 @@ message SummaryRequest{
     string Operation = 6;
 }
 
+message Podowner {
+  string Ref = 1;
+  string Name = 2;
+  string Namespace = 3;
+}
```
wazir-ahmed:
```suggestion
message Workload {
  string Type = 1;
  string Name = 2;
}
```

wazir-ahmed (changes requested):

src/types/observability.go
```diff
@@ -94,39 +94,43 @@ type NetworkSummary struct {
...
+	UpdatedTime    int64    `json:"UpdatedTime,omitempty"`
+	WorkspaceId    int32    `json:"WorkspaceId,omitempty"`
+	BindPort       string   `json:"BindPort,omitempty"`
+	BindAddress    string   `json:"BindAddress,omitempty"`
+	Severity       string   `json:"Severity,omitempty"`
+	Tags           string   `json:"Tags,omitempty"`
+	Message        string   `json:"Message,omitempty"`
+	Enforcer       string   `json:"Enforcer,omitempty"`
+	PolicyName     string   `json:"PolicyName,omitempty"`
+	Workload       Workload `json:"Owner,omitempty"`
```
wazir-ahmed:
```suggestion
	Workload       Workload `json:"Workload,omitempty"`
```

src/protobuf/v1/publisher/publisher.proto
```diff
@@ -29,24 +34,24 @@ message SummaryResponse{
...
-    int64 UpdatedTime = 23;
-    int32 WorkspaceId = 24;
-    string BindPort = 25;
-    string BindAddress = 26;
-    string Severity = 27;
-    string Tags = 28;
-    string Message = 29;
-    string Enforcer = 30;
-    string PolicyName = 31;
+    string Source = 12;
```
wazir-ahmed: In ProtoBuf, we should not re-assign the numbers to different fields. That will make the application incompatible with the older clients. We should always add new values at the end, incrementing last field number. 

To deprecate an existing field, either you can remove it from the definition or comment it out or mark the field as `deprecated=true`.
```proto
message MyMessage {
    // string my_field = 1
    string my_new_field = 2;
}

message MyMessage {
    string my_new_field = 2;
}

message MyMessage {
    string my_field = 1 [deprecated=true];
    string my_new_field = 2;
}
```
Marking it as `deprecated=true` seems to be a better choice.

src/observability/consumer.go
```diff
@@ -83,10 +83,6 @@ func validateSummaryRequest(consumer *SummaryConsumer, summary types.SystemSumma
 		return false
 	}
 
-	if consumer.Deployment != "" && consumer.Deployment != summary.Deployment {
```
wazir-ahmed: If we're removing the `Deployment` field, shouldn't we use `Workload.Name` in all other places where we're use `Deployment` field? Check if this removal would affect `karmor summary` in anyway.

wazir-ahmed (changes requested):

src/observability/consumer.go
wazir-ahmed: @seswarrajan WDYT? Are we depending on `Deployment` field in any case?

seswarrajan: Yes... We are dependent on deployment field. I suggest not removing this

src/observability/kubearmor.go
```diff
@@ -229,7 +229,6 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
 		ContainerName: req.ContainerName,
 		ClusterName:   req.ClusterName,
 		Labels:        req.Label,
-		Deployment:    req.DeployName,
```
seswarrajan: @AmanKumar2696 Do not remove this. This is gRPC incoming request field

src/observability/kubearmor.go
```diff
@@ -242,7 +241,6 @@ func GetKubearmorSummaryData(req *opb.Request) ([]types.SysObsProcFileData, []ty
 			podInfo.ContainerName = ss.ContainerName
 			podInfo.Labels = ss.Labels
 			podInfo.Namespace = ss.NamespaceName
-			podInfo.DeployName = ss.Deployment
```
seswarrajan: @AmanKumar2696 Do not remove this. This is gRPC incoming request field

seswarrajan (changes requested):

src/observability/consumer.go
```diff
@@ -87,6 +87,9 @@ func validateSummaryRequest(consumer *SummaryConsumer, summary types.SystemSumma
 		return false
 	}
 
+	if consumer.Labels != "" && consumer.Labels != summary.Labels {
+		return false
+	}
```
wazir-ahmed:
```suggestion
```

src/libs/dbHandler.go
```diff
@@ -457,7 +457,6 @@ func upsertSysSummarySQL(db *sql.DB, summary types.SystemSummary, timeCount type
 		summary.PodName,
 		summary.Operation,
 		summary.Labels,
-		summary.Deployment,
```
wazir-ahmed: Let's keep the deployment related code in DB handlers. Please undo `summary.Deployment` changes from this file.

src/observability/summarizer.go
```diff
@@ -124,14 +122,8 @@ func convertSysLogToSysSummaryMap(syslogs []*pb.Alert) {
 		sysSummary.Message = syslog.Message
 		sysSummary.Severity = syslog.Severity
 		sysSummary.PolicyName = syslog.PolicyName
-		sysSummary.Deployment = ""
```
wazir-ahmed, line 127: For backward compatibility, assign `sysSummary.Deployment = syslog.Owner.Name`.

src/observability/consumer.go
```diff
@@ -109,29 +112,28 @@ func validateSummaryRequest(consumer *SummaryConsumer, summary types.SystemSumma
...
+		DestNamespace: summary.DestNamespace,
+		DestLabels:    summary.DestLabels,
+		NwType:        summary.NwType,
+		IP:            summary.IP,
+		Port:          summary.Port,
+		Protocol:      summary.Protocol,
+		Action:        summary.Action,
+		Count:         summary.Count,
+		UpdatedTime:   summary.UpdatedTime,
+		WorkspaceId:   summary.WorkspaceId,
```
wazir-ahmed, line 135: We're missing `Summary.Workload` here.

wazir-ahmed (changes requested):

src/protobuf/v1/publisher/publisher.proto
wazir-ahmed: @AmanKumar2696 This is not not yet addressed properly. Please mark the old field as deprecated.

AmanKumar2696: For backward compatibility, this field wasn't marked as deprecated.

src/libs/dbHandler.go
```diff
@@ -502,7 +502,7 @@ func upsertSysSummarySQL(db *sql.DB, summary types.SystemSummary, timeCount type
 			summary.PodName,
 			summary.Operation,
 			summary.Labels,
-			summary.Deployment,
+			summary.Destination,
```
wazir-ahmed: Please revert this back to `Deployment`

wazir-ahmed (changes requested):

src/protobuf/v1/publisher/publisher.proto
wazir-ahmed: Marking it as deprecated will not affect the backward compatibility.

wazir-ahmed (approved):
Looks good to me :+1:

seswarrajan (approved):
LGTM as well.

Issue #715, opened 2023-04-25
feat: discover lenient whitelisting policies - Crown Jewel


Ankurk99:
**Description**
The current discovered security policies are not usable. It's because we are discovering least-permissive policies (for zero trust) by default and if the discovered policy miss any important binary, it can makes the application unusable or even crash.
The aim is to discover lenient whitelisting policies - starting with the policies based on mount points.

**Working**
- Get all the used mounted mount points from the deployment
- Check if the mount point is being accessed or not. If not, then deny access to that particular mount point.

cc: @nyrahul

PR #739, opened 2023-06-05
add support for crown jewel policies

Ankurk99:
**Description**

Add support for generating lenient policies protecting sensitive assets (mount points here)

**Problem**
Currently we were discovering least-permissive security policies using Discovery Engine. Ideally these policies were designed to get us a Zero-trust security posture by only allowing the binaries which are essential for running any particular application. But the issue with these types of policies is that it's very difficult to actually create an exhaustive policy which contains all the necessary binaries allowed. If we miss even a single important binary, that would have caused the whole application to crash in worst case.

**Solution**
The aim of this PR is to identify the crown-jewels or the list of important assets which when protected can give us a fairly good security posture for the whole application. 

**PR changes**
This PR identifies the paths mounted by any application and check if they are actually being used. We then create a "Crown-jewel" lenient policy which only allows the access of the particular mount paths by a particular binary based on the actual usage and deny access from others.

Example of a Crown jewel policy:
```
 - apiVersion: v1
  kind: KubeArmorPolicy
  metadata:
    name: autopol-assets-vault
    namespace: default
  spec:
    action: Allow
    file:
      matchDirectories:
      - action: Block
        dir: /vault/data/
        recursive: true
      - dir: /vault/data/
        fromSource:
        - path: /bin/vault
        recursive: true
      - dir: /
        recursive: true
      - dir: /vault/config/
        recursive: true
      - action: Block
        dir: /home/vault/
        recursive: true
      - dir: /home/vault/
        fromSource:
        - path: /bin/sh
        recursive: true
    message: Sensitive assets and process control policy
    network: {}
    process:
      matchPaths:
      - path: /bin/sh
      - path: /bin/vault
      - path: /bin/busybox
    selector:
      matchLabels:
        app.kubernetes.io/instance: vault
        app.kubernetes.io/name: vault
        component: server
        helm.sh/chart: vault-0.24.1
    severity: 7
```

Here in the above policy, access to the dir:  `/home/vault/` is only allowed by `/bin/sh` and `/vault/data/` by `/bin/vault`.  The other mount path `/vault/config/` is not being used, so it's set to `Block`. As Vault is using Alpine image, that's why we see the process `/bin/busybox` being used (will be a symlink to system packages).

ref: https://github.com/accuknox/discovery-engine/issues/715

### DIFF
--- a/src/crownjewel/crownjewel.go
+++ b/src/crownjewel/crownjewel.go
@@ -1,0 +1,392 @@
+package crownjewel
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/accuknox/auto-policy-discovery/src/cluster"
+	"github.com/accuknox/auto-policy-discovery/src/common"
+	"github.com/accuknox/auto-policy-discovery/src/config"
+	cfg "github.com/accuknox/auto-policy-discovery/src/config"
+	"github.com/accuknox/auto-policy-discovery/src/libs"
+	logger "github.com/accuknox/auto-policy-discovery/src/logging"
+	obs "github.com/accuknox/auto-policy-discovery/src/observability"
+	opb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/observability"
+	"github.com/accuknox/auto-policy-discovery/src/systempolicy"
+	"github.com/accuknox/auto-policy-discovery/src/types"
+	"github.com/robfig/cron"
+	"github.com/rs/zerolog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var log *zerolog.Logger
+
+// CrownjewelCronJob for cron job
+var CrownjewelCronJob *cron.Cron
+
+var CrownjewelStopChan chan struct{}
+
+// CrownjewelWorkerStatus global worker
+var CrownjewelWorkerStatus string
+
+// const values
+const (
+	// operation mode
+	opModeNoop    = 0
+	opModeCronjob = 1
+
+	// status
+	statusRunning = "running"
+	statusIdle    = "idle"
+)
+
+// init Function
+func init() {
+	log = logger.GetInstance()
+	CrownjewelWorkerStatus = statusIdle
+	CrownjewelStopChan = make(chan struct{})
+}
+
+// StartCrownjewelWorker starts the crown jewel worker (run once or as a cronjob)
+func StartCrownjewelWorker() {
+	if CrownjewelWorkerStatus != statusIdle {
+		log.Info().Msg("There is no idle Crown jewel policy worker")
+		return
+	}
+	if cfg.GetCfgCrownjewelOperationMode() == opModeNoop { // Do not run the operation
+		log.Info().Msg("Crown jewel operation mode is NOOP... ")
+	} else if cfg.GetCfgCrownjewelOperationMode() == opModeCronjob { // every time intervals
+		log.Info().Msg("Crown jewel policy cron job started")
+		CrownjewelPolicyMain()
+		StartCrownjewelCronJob()
+	} else { // one-time generation
+		CrownjewelPolicyMain()
+		log.Info().Msgf("Crown jewel policy onetime job done")
+	}
+}
+
+// StartCrownjewelCronJob starts the cronjob
+func StartCrownjewelCronJob() {
+	// init cron job
+	CrownjewelCronJob = cron.New()
+	err := CrownjewelCronJob.AddFunc(cfg.GetCfgCrownjewelCronJobTime(), CrownjewelPolicyMain)
+	if err != nil {
+		log.Error().Msg(err.Error())
+		return
+	}
+	CrownjewelCronJob.Start()
+}
+
+// StopCrownjewelCronJob stops the cronjob
+func StopCrownjewelCronJob() {
+	if CrownjewelCronJob != nil {
+		log.Info().Msg("Got a signal to terminate the auto system policy discovery")
+
+		CrownjewelStopChan = make(chan struct{})
+
+		close(CrownjewelStopChan)
+
+		CrownjewelCronJob.Stop() // Stop the scheduler (does not stop any jobs already running).
+
+		CrownjewelCronJob = nil
+	}
+}
+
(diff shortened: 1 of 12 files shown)


src/types/policyData.go
```diff
@@ -244,6 +244,7 @@ type KnoxMatchPaths struct {
 	ReadOnly   bool             `json:"readOnly,omitempty" yaml:"readOnly,omitempty"`
 	OwnerOnly  bool             `json:"ownerOnly,omitempty" yaml:"ownerOnly,omitempty"`
 	FromSource []KnoxFromSource `json:"fromSource,omitempty" yaml:"fromSource,omitempty"`
+	Action     string           `json:"action,omitempty"`
```
Ankurk99, line 250: The generated policies doesn't have the "Action" field and we were importing the type from KubeArmor. Have added it for handling action for crown jewel policies. Hope it doesn't have any bad implications.
cc: @nyrahul @vishnusomank

src/crownjewel/crownjewel.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+	labelsString := strings.Join(combinedLabels, ",")
+
+	// create policy name
+	name := strconv.FormatUint(uint64(common.HashInt(ns+clustername+cname+labelsString)), 10)
+	return types.KnoxSystemPolicy{
+		APIVersion: "v1",
+		Kind:       "KubeArmorPolicy",
+		Metadata: map[string]string{
+			"name":      "autopol-sensitive-" + name,
```
Ankurk99, line 390: Any changes required in policy name?
@nyrahul

src/crownjewel/crownjewel.go
```diff
@@ -0,0 +1,432 @@
...
+		CrownjewelCronJob.Stop() // Stop the scheduler (does not stop any jobs already running).
+
+		CrownjewelCronJob = nil
+	}
+}
+
+// Create Crown Jewel Policy based on K8s object type
+func CrownjewelPolicyMain() {
+	client := cluster.ConnectK8sClient()
+	deployments, err := client.AppsV1().Deployments("").List(context.Background(), metav1.ListOptions{})
```
vishnusomank: can we use the `cluster` package function GetDeploymentsFromK8sClient() here instead of calling the client-go? Since we only need the name, namespace, and labels it will suffice. Also, we can remove the multiple for loops if we append the sts,rs,ds values to a common variable

vishnusomank (changes requested):

src/crownjewel/crownjewel.go
Ankurk99: @vishnusomank comment addressed. Thanks for the suggestion.

src/crownjewel/crownjewel.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+	CrownjewelCronJob.Start()
+}
+
+// StopCrownjewelCronJob stops the cronjob
+func StopCrownjewelCronJob() {
+	if CrownjewelCronJob != nil {
+		log.Info().Msg("Got a signal to terminate the auto system policy discovery")
+
+		CrownjewelStopChan = make(chan struct{})
```
rajaSahil, line 88: I can see this channel also getting initialized through init, can you explain the flow here?

src/crownjewel/crownjewel.go
```diff
@@ -0,0 +1,392 @@
...
+		LabelSelector: labels,
+	})
+	if err != nil {
+		log.Warn().Msg(err.Error())
+	}
+
+	for _, pod := range podList.Items {
+		for _, container := range pod.Spec.Containers {
+			sumResp, err := obs.GetSummaryData(&opb.Request{
+				PodName:       pod.Name,
```
rajaSahil, line 212: Why are we using pod name?
And how mount paths from observability data is utilized?

Ankurk99: > Why are we using pod name?

We are checking the resource to be pods (instead of deployment/statefulset etc.) because pod labels contains all the labels for that resource which might be missing in other workload (say statefulset for eg.).
Here below is an example of a vault statefulset:

Pod Labels:
```
Labels:     app.kubernetes.io/instance=vault
                  app.kubernetes.io/name=vault
                  component=server
                  controller-revision-hash=vault-5f4c59685d
                  helm.sh/chart=vault-0.24.1
                  statefulset.kubernetes.io/pod-name=vault-0
```
Statefulset labels:
```
  Labels:           app.kubernetes.io/instance=vault
                    app.kubernetes.io/name=vault
                    component=server
                    helm.sh/chart=vault-0.24.1```

Ankurk99:
The code implementation can be summarized as below:

- func getProcessList() : Get the list of running processes from observability data and store it in an array processList[]
- func getVolumeMountPaths(): Get all the volume path from the k8s cluster (looking for pods matching the labels) and store it in array mountPaths[]
- func usedMountPath(): Get the list of used mount paths from observability data and store the info together in sumResponses[] string with fromSource make(map[string]string)
- func accessedMountPaths(): Match used mounts paths with actually accessed mount paths. Here we compare the mount points found in getVolumeMountPaths and usedMountPaths, if they match then keep them otherwise ignore those mount paths.

All the required info about the mount points which are mounted and actually being used are used to create a crown jewel policy using getCrownjewelPolicy() func.
- func getCrownjewelPolicy() calls createCrownjewelPolicy()
- func createCrownjewelPolicy(): Assigns Actions (for eg: block Allow for matching mount paths and Block by default) and ignores duplicate fromSource values. It then calls buildSystemPolicy() func to generate the crown jewel policy.
- func buildSystemPolicy(): has the template to create the policy. Currently, severity and message are set by default but they can be assigned from the func arguments.

- func getFilteredPolicy(): filters to check the namespaces to be ignored and calls systempolicy.UpdateSysPolicies(policies)
- func systempolicy.UpdateSysPolicies(): Inserts the system policies to DB (sqlite or MYSQL)
- func WriteSystemPoliciesToFile(): all the crown jewel policies are saved to a file named "kubearmor_policies_sensitive"

The logic for updating the crown jewel policies to the DB is the same as was for other system policies.

src/crownjewel/crownjewel.go
Ankurk99: >  And how mount paths from observability data is utilized?

PTAL at https://github.com/accuknox/discovery-engine/pull/739#issuecomment-1741382140, if anything is unclear please comment.

PR #746, opened 2023-06-15
[fix](template-version) Adding ability to specify policy-template version

vishnusomank:
This PR contains the following changes 

- Updated policy template download logic to get a specific version of PT or the latest based on config flag
- Use existing cache if a specific version is specified using the config file

**Purpose of PR?**:

The discovery engine's recommendation feature heavily relies on the policy-template release branch. If there is an update on the policy template release older/stable versions of Discovery Engine might break. To avoid this we are introducing the ability to specify the version of the policy template release to be used in the Discovery Engine

Fixes #

**Does this PR introduce a breaking change?** No

**If the changes in this PR are manually verified, list down the scenarios covered:**:
- added policy-template version 'v0.2.2' to the config file
- omitted policy-template version to check if new policy-template releases are getting downloaded 

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Bug fix. Fixes #<issue number>
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests

### DIFF
--- a/src/conf/local-file.yaml
+++ b/src/conf/local-file.yaml
@@ -119,6 +119,7 @@ recommend:
   operation-mode: 1                       # 1: cronjob | 2: one-time-job
   cron-job-time-interval: "1h0m00s"       # format: XhYmZs
   recommend-host-policy: true
+  template-version: "v0.2.2"              # policy template version to be used for recommendation (keep empty to fetches latest)
 
 # license
 license:
--- a/src/conf/local.yaml
+++ b/src/conf/local.yaml
@@ -83,6 +83,7 @@ recommend:
   operation-mode: 1                       # 1: cronjob | 2: one-time-job
   cron-job-time-interval: "1h0m00s"       # format: XhYmZs
   recommend-host-policy: true
+  template-version: "v0.2.1"
 
 # license
 license:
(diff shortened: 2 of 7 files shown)


src/conf/local-file.yaml
```diff
@@ -119,6 +119,7 @@ recommend:
   operation-mode: 1                       # 1: cronjob | 2: one-time-job
   cron-job-time-interval: "1h0m00s"       # format: XhYmZs
   recommend-host-policy: true
+  template-version: "v0.2.2"              # policy template version to be used for recommendation (keep empty to fetches latest)
```
nyrahul, line 122: Is it possible for me to say `template-version: "latest"` to download/use the latest one?

nyrahul (commented):
I feel overall we are handling this the wrong way. We should ideally create a STABLE release and use those. The STABLE release can change from time to time. Internal developers can still set it to `latest` and use it.

The problem with the current approach is, once we have shipped a particular version to the user there is no way for us to change it. With STABLE mechanism we can still change it.

src/conf/local-file.yaml
vishnusomank: yes @nyrahul we can keep it empty to download the latest.

wazir-ahmed (approved):
LGTM

PR #755, opened 2023-06-24
feat(observability): Aggregate summary data before publishing to stream

vishnusomank:
**Purpose of PR?**:

The summary observability publishes all events without aggregating. This leads to a huge number of messages getting published in the stream. This PR reduces the number of summary information being published to the stream by aggregating the destination path and reducing the data by replacing the destination with the aggregated path and updating the count and ipdated_time.

Fixes _**JIRA CNAPP-6011**_

**Does this PR introduce a breaking change?** 
_**NO**_

**Checklist:**
- [x] New feature (non-breaking change which adds functionality)
- [ ] Commit has unit tests

### DIFF
--- a/src/observability/publisher.go
+++ b/src/observability/publisher.go
@@ -1,7 +1,9 @@
 package observability
 
 import (
+	"strings"
 	"sync"
+	"time"
 
 	"github.com/accuknox/auto-policy-discovery/src/common"
 	cfg "github.com/accuknox/auto-policy-discovery/src/config"
@@ -27,8 +29,15 @@ func ProcessSystemSummary() {
 
 	if cfg.GetCfgPublisherEnable() {
 		PublisherMutex.Lock()
+		timeVar := time.Now()
+		tempSummarizerMap = getAggregatedSummaryMap(tempSummarizerMap)
+
+		log.Info().Msgf("Events to publish after aggregation: [%v]", len(tempSummarizerMap))
+		count := 0
+
 		// publish summary map in GRPC
 		for ss, sstc := range tempSummarizerMap {
+			count++
 			var locSummary types.SystemSummary = ss
 
 			// update count/time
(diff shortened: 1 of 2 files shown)


src/observability/publisher.go
```diff
@@ -41,7 +49,60 @@ func ProcessSystemSummary() {
...
+	aggregatedFilePaths := common.AggregatePaths(fileArr)
+	tmpFileSumMap := make(map[string]map[types.SystemSummary]types.SysSummaryTimeCount)
+
+	log.Info().Msgf("Aggregated file paths: [%v]", len(aggregatedFilePaths))
+	for ss, sstc := range fileSumMap {
+		localSs := ss
+		localSstc := sstc
+		localMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
+		for _, path := range aggregatedFilePaths {
+			if strings.HasPrefix(ss.Destination, path.Path) {
```
nyrahul: consider the situation where `ss.Destination = /usr/local/bin/xyz` and `path.Path = /usr/local/b` ... This will match whereas it should not.

src/observability/publisher.go
```diff
@@ -41,7 +49,60 @@ func ProcessSystemSummary() {
...
+				localSs.Destination = path.Path
+			}
+		}
+		key := localSs.PodName + "_" + localSs.ContainerName + "_" + localSs.Source + "_" + localSs.Destination
+		data, ok := tmpFileSumMap[key]
```
nyrahul: Can the len(data) be greater than 1 here? If yes, then I think the logic below is incorrect.

src/observability/publisher.go
```diff
@@ -41,7 +49,60 @@ func ProcessSystemSummary() {
...
+	for ss, sstc := range fileSumMap {
+		localSs := ss
+		localSstc := sstc
+		localMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
+		for _, path := range aggregatedFilePaths {
+			if strings.HasPrefix(ss.Destination, path.Path) {
+				localSs.Destination = path.Path
+			}
+		}
+		key := localSs.PodName + "_" + localSs.ContainerName + "_" + localSs.Source + "_" + localSs.Destination
```
nyrahul: We need namespace to be added to the key as well.

It is possible to have same podname + containername + source + destination across different namespaces.

nyrahul (changes requested):
The computational efficiency of this logic is not good. This will cause CPU spikes during processing. Please check other comments inline.

src/observability/publisher.go
vishnusomank: I've added additional checks to make the matching more efficient

src/observability/publisher.go
vishnusomank: No the len(data) will always be 1, but I've updated the logic to not create additional maps

src/observability/publisher.go
```diff
@@ -41,7 +50,51 @@ func ProcessSystemSummary() {
...
+	aggregatedFilePaths := common.AggregatePaths(fileArr)
+
+	log.Info().Msgf("Aggregated file paths: [%v]", len(aggregatedFilePaths))
+	for ss, sstc := range fileSumMap {
+		for _, path := range aggregatedFilePaths {
+			if strings.HasPrefix(ss.Destination, path.Path) && (len(ss.Destination) == len(path.Path) || ss.Destination[len(strings.TrimSuffix(path.Path, "/"))] == '/') {
+				ss.Destination = path.Path
+				break
+			}
+		}
```
seswarrajan, line 85:
```suggestion
		    delete(fileSumMap, ss)
		}
```

src/observability/publisher.go
```diff
@@ -41,7 +50,51 @@ func ProcessSystemSummary() {
...
 	}
 	ProcessSystemSummaryWg.Wait()
 }
+
+func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) map[types.SystemSummary]types.SysSummaryTimeCount {
+
+	var fileArr []string
+	log.Info().Msgf("Events before aggregation: [%v]", len(tempSummarizerMap))
+
+	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
```
seswarrajan, line 64: @vishnusomank Just a suggestion, call make of maps from some common init code and use it here.

src/observability/publisher.go
```diff
@@ -41,7 +50,51 @@ func ProcessSystemSummary() {
...
+	updatedSummarizerMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
```
seswarrajan, line 65: Where do we clear this updatedSummarizerMap ?

seswarrajan (changes requested):

src/observability/publisher.go
vishnusomank: Since this map is local to getAggregatedSummaryMap(), I assume that every time the function gets called it'll automatically get recreated. Do you think we need to clear it separately?

src/observability/publisher.go
vishnusomank: Since the iteration is happening only via one loop do we need to clear elements from the map? I thought adding this delete() might increase the time complexity of the loop. What do you suggest @seswarrajan ?

seswarrajan (approved):
LGTM

nyrahul (approved):

stefin9898 (approved):
LGTM

src/observability/publisher.go
```diff
@@ -41,7 +50,51 @@ func ProcessSystemSummary() {
...
+	for tss, tsstc := range tempSummarizerMap {
+		if tss.Operation == types.FileOperation {
+			fileArr = append(fileArr, tss.Destination)
```
wazir-ahmed, line 68: We're using a common file array here and appending file paths from different pods and aggregating paths across all the pods. This is not correct. 

We should only run the aggregate path logic on events from a single pod/workload.

wazir-ahmed (changes requested):

PR #760, opened 2023-06-27
fix(observability): Updated file aggregation logic based on pod_name

vishnusomank:
**Purpose of PR?**:

The current file aggregation logic for observability summary will aggregate file paths without taking into consideration of the pod details. This PR will make the file path aggregation per pod leading to a much more accurate summary detail. 

Fixes #

**Does this PR introduce a breaking change?**
**_NO_**

**Checklist:**
- [x] Bug fix. 
- [x] Commit has unit tests

### DIFF
--- a/src/observability/publisher.go
+++ b/src/observability/publisher.go
@@ -29,72 +29,73 @@ func ProcessSystemSummary() {
 
 	if cfg.GetCfgPublisherEnable() {
 		PublisherMutex.Lock()
-		timeVar := time.Now()
-		tempSummarizerMap = getAggregatedSummaryMap(tempSummarizerMap)
+		initTime := time.Now()
+
+		aggregateSummaryMap(tempSummarizerMap)
 
 		log.Info().Msgf("Events to publish after aggregation: [%v]", len(tempSummarizerMap))
 		count := 0
 
 		// publish summary map in GRPC
-		for ss, sstc := range tempSummarizerMap {
+		for sysSummary, summaryTimeCount := range tempSummarizerMap {
 			count++
-			var locSummary types.SystemSummary = ss
+			var locSummary types.SystemSummary = sysSummary
 
 			// update count/time
-			locSummary.Count = sstc.Count
-			locSummary.UpdatedTime = sstc.UpdatedTime
+			locSummary.Count = summaryTimeCount.Count
+			locSummary.UpdatedTime = summaryTimeCount.UpdatedTime
 
 			// publish data to feeder grpc
 			SysSummary.Publish(&locSummary)
 
 			// clear each published entry from data map
-			delete(tempSummarizerMap, ss)
+			delete(tempSummarizerMap, sysSummary)
 		}
-		log.Info().Msgf("Published %v events in %v", count, time.Since(timeVar))
+		log.Info().Msgf("Published %v events in %v", count, time.Since(initTime))
 		PublisherMutex.Unlock()
 	}
 	ProcessSystemSummaryWg.Wait()
 }
 
-func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) map[types.SystemSummary]types.SysSummaryTimeCount {
+func aggregateSummaryMap(summaryMap map[types.SystemSummary]types.SysSummaryTimeCount) {
 
-	var fileArr []string
-	log.Info().Msgf("Events before aggregation: [%v]", len(tempSummarizerMap))
+	podFilePaths := make(map[string][]string)
+	aggPodFilePaths := make(map[string][]common.SysPath)
+	fileSummarizerMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
 
-	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
-	updatedSummarizerMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
-	for tss, tsstc := range tempSummarizerMap {
-		if tss.Operation == types.FileOperation {
-			fileArr = append(fileArr, tss.Destination)
-			fileSumMap[tss] = tsstc
-		} else {
-			updatedSummarizerMap[tss] = tsstc
+	log.Info().Msgf("Events before aggregation: [%v]", len(summaryMap))
+	for sysSummary, summaryTimeCount := range summaryMap {
+		if sysSummary.Operation == types.FileOperation {
+			key := sysSummary.PodName + "_" + sysSummary.Source
+			podFilePaths[key] = append(podFilePaths[key], sysSummary.Destination)
+			fileSummarizerMap[sysSummary] = summaryTimeCount
+			delete(summaryMap, sysSummary)
 		}
 	}
-	tempSummarizerMap = updatedSummarizerMap
 
-	aggregatedFilePaths := common.AggregatePaths(fileArr)
+	for key, files := range podFilePaths {
+		aggPodFilePaths[key] = common.AggregatePaths(files)
+		log.Info().Msgf("Got %v aggregated file paths for key [%v]", len(aggPodFilePaths[key]), key)
+	}
 
-	log.Info().Msgf("Aggregated file paths: [%v]", len(aggregatedFilePaths))
-	for ss, sstc := range fileSumMap {
-		for _, path := range aggregatedFilePaths {
-			if strings.HasPrefix(ss.Destination, path.Path) && (len(ss.Destination) == len(path.Path) || ss.Destination[len(strings.TrimSuffix(path.Path, "/"))] == '/') {
-				ss.Destination = path.Path
+	for sysSummary, summaryTimeCount := range fileSummarizerMap {
+		key := sysSummary.PodName + "_" + sysSummary.Source
+		files := aggPodFilePaths[key]
+		for _, path := range files {
+			if strings.HasPrefix(sysSummary.Destination, path.Path) && (len(sysSummary.Destination) == len(path.Path) || sysSummary.Destination[len(strings.TrimSuffix(path.Path, "/"))] == '/') {
+				sysSummary.Destination = path.Path
 				break
 			}
 		}
-
-		if existingSstc, ok := tempSummarizerMap[ss]; ok {
-			existingSstc.Count += sstc.Count
-			if sstc.UpdatedTime > existingSstc.UpdatedTime {
-				existingSstc.UpdatedTime = sstc.UpdatedTime
+		if timeCount, ok := summaryMap[sysSummary]; ok {
+			timeCount.Count += summaryTimeCount.Count
+			if summaryTimeCount.UpdatedTime > timeCount.UpdatedTime {
+				timeCount.UpdatedTime = summaryTimeCount.UpdatedTime
 			}
-			tempSummarizerMap[ss] = existingSstc
+			summaryMap[sysSummary] = timeCount
 		} else {
-			tempSummarizerMap[ss] = sstc
+			summaryMap[sysSummary] = summaryTimeCount
 		}
 
 	}
-
-	return tempSummarizerMap
 }
--- a/src/observability/publisher_test.go
+++ b/src/observability/publisher_test.go
@@ -1,0 +1,104 @@
+package observability
+
+import (
+	"encoding/json"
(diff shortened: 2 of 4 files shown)


src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
...
 
-	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
 	updatedSummarizerMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
 	for tss, tsstc := range tempSummarizerMap {
 		if tss.Operation == types.FileOperation {
-			fileArr = append(fileArr, tss.Destination)
+			destination := []string{}
+			destination = append(destination, fileArr[tss.PodName]...)
+			destination = append(destination, tss.Destination)
+			fileArr[tss.PodName] = destination
```
wazir-ahmed:
```suggestion
			fileArr[tss.PodName] = append(fileArr[tss.PodName], tss.Destination)
```

src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
...
 	log.Info().Msgf("Events before aggregation: [%v]", len(tempSummarizerMap))
 
-	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
 	updatedSummarizerMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
 	for tss, tsstc := range tempSummarizerMap {
 		if tss.Operation == types.FileOperation {
-			fileArr = append(fileArr, tss.Destination)
+			destination := []string{}
+			destination = append(destination, fileArr[tss.PodName]...)
+			destination = append(destination, tss.Destination)
```
seswarrajan: @vishnusomank I do not understand this logic. Why do we add Podname and Destination both to same string slice ?
What is tss.Destination ?

vishnusomank: We are not adding pod_name to the string slice. We are creating a map[pod_name][]string to store the file paths of each pod before aggregating them. This will make the aggregation per pod and lead to more accurate summary information
cc: @wazir-ahmed

src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
...
 
-	var fileArr []string
+	fileArr := make(map[string][]string)
+	fileArrTmp := make(map[string][]common.SysPath)
+	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
+
 	log.Info().Msgf("Events before aggregation: [%v]", len(tempSummarizerMap))
 
-	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
 	updatedSummarizerMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
```
wazir-ahmed: There is no need for this variable `updatedSummarizerMap`. Instead of copying non-file events to this map and then reassigning `tempSummarizerMap` with this variable,  we could just delete the file events from `tempSummarizerMap`. We can avoid allocating additional memory.

src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
 
 func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) map[types.SystemSummary]types.SysSummaryTimeCount {
```
wazir-ahmed:
```suggestion
func aggregateFileSummary(summary map[types.SystemSummary]types.SysSummaryTimeCount)  {
```
Since this function modify the input, there no need for a return type.

src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
...
 
-	var fileArr []string
+	fileArr := make(map[string][]string)
+	fileArrTmp := make(map[string][]common.SysPath)
```
wazir-ahmed, line 62:
```suggestion
	filePaths := make(map[string][]string)
	aggFilePaths := make(map[string][]common.SysPath)
```

src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
...
+	fileSumMap := make(map[types.SystemSummary]types.SysSummaryTimeCount)
```
wazir-ahmed:
```suggestion
	filePaths := make(map[string][]string)
	aggFilePaths := make(map[string][]common.SysPath)
	fileSummary := make(map[types.SystemSummary]types.SysSummaryTimeCount)
```

src/observability/publisher.go
```diff
@@ -58,32 +58,40 @@ func ProcessSystemSummary() {
...
 		} else {
 			updatedSummarizerMap[tss] = tsstc
 		}
 	}
 	tempSummarizerMap = updatedSummarizerMap
 
-	aggregatedFilePaths := common.AggregatePaths(fileArr)
+	for key, files := range fileArr {
+		aggregatedFilePaths := common.AggregatePaths(files)
+		fileArrTmp[key] = aggregatedFilePaths
```
wazir-ahmed:
```suggestion
		fileArrTmp[key] = common.AggregatePaths(files)

File: src/observability/publisher.go [hunk 8]
```diff
@@ -29,72 +29,71 @@ func ProcessSystemSummary() {
...
 	ProcessSystemSummaryWg.Wait()
 }
 
-func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) map[types.SystemSummary]types.SysSummaryTimeCount {
+func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) {
 
-	var fileArr []string
-	log.Info().Msgf("Events before aggregation: [%v]", len(tempSummarizerMap))
+	podFilePaths := make(map[string][]string)
+	tempPodFilePaths := make(map[string][]common.SysPath)
```
wazir-ahmed:
```suggestion
	aggPodFilePaths := make(map[string][]common.SysPath)
```

src/observability/publisher.go
```diff
@@ -29,72 +29,71 @@ func ProcessSystemSummary() {
...
 		}
-		log.Info().Msgf("Published %v events in %v", count, time.Since(timeVar))
+		log.Info().Msgf("Published %v events in %v", count, time.Since(initTime))
 		PublisherMutex.Unlock()
 	}
 	ProcessSystemSummaryWg.Wait()
 }
 
-func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) map[types.SystemSummary]types.SysSummaryTimeCount {
+func getAggregatedSummaryMap(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount) {
```
wazir-ahmed:
```suggestion
func aggregateSummaryMap(summaryMap map[types.SystemSummary]types.SysSummaryTimeCount) {
```

src/observability/publisher.go
```diff
@@ -29,72 +29,71 @@ func ProcessSystemSummary() {
...
+				sysSummary.Destination = path.Path
 				break
 			}
 		}
-
-		if existingSstc, ok := tempSummarizerMap[ss]; ok {
-			existingSstc.Count += sstc.Count
-			if sstc.UpdatedTime > existingSstc.UpdatedTime {
-				existingSstc.UpdatedTime = sstc.UpdatedTime
+		if existingSstc, ok := tempSummarizerMap[sysSummary]; ok {
```
wazir-ahmed:
```suggestion
		if timeCount, ok := tempSummarizerMap[sysSummary]; ok {
```

src/observability/publisher_test.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// makeSystemSummaryData reads system summary data from a file 
+func makeSystemSummaryData(tempSummarizerMap map[types.SystemSummary]types.SysSummaryTimeCount)  {
+
+	var systemSummary types.SystemSummary
+	var systemSummaryCount types.SysSummaryTimeCount
+
+	// Open the file
+	file, err := os.Open("summary-data.json")
```
wazir-ahmed: 1. The input is not a `JSON`. We should take a dump of summaryMap in `JSON` format and use it as the input.
2. The test case doesn't contain the expected output. We're just hardcoding the `len(summaryMap)` before & after aggregation which is not the correct way to handle it.

src/observability/input-summary.json
```diff
@@ -0,0 +1,36 @@
+{"ClusterName":"default","Namespace":"accuknox-agents","ContainerName":"policy-enforcement-agent","ContainerImage":"public.ecr.aws/k9v9d5v2/policy-enforcement-agent:v0.1.0@sha256:005c1fde3ff8a667f3ac7540c5c011c752a7e3aaa2c89aa335703289ed8d80f8","ContainerID":"9af07242b00507a9e7ec1c017a616f6be766f15aa4e6d25133a672a248eb8a5e","PodName":"policy-enforcement-agent-7946b64dfb-lc95w","Operation":"Network","Labels":"app=policy-enforcement-agent","Deployment":"policy-enforcement-agent","Source":"/home/pea/main","Type":"egress","IP":"3.130.218.162","Port":443,"Protocol":"TCP","Action":"Allow","Workload":{"Type":"Deployment","Name":"policy-enforcement-agent"}}
+{"Count":4,"UpdatedTime":1688108538}
```
wazir-ahmed: For the test, we don't need all the summary information. Can we change the input/output format to the following, so that it would be easy to add more test cases.
```suggestion
{
    "test1": [
        {
            "summary": {"PodName":"shared-informer-agent-5cd8b49c55-xhn4z", "Operation":"File", "Source":"/home/sia/main", "Resource":"/var/run/secrets/kubernetes.io/serviceaccount/token"},
            "timeCount": { "Count":4, "UpdatedTime":1688108538}
        },
        {
            "summary": {},
            "timeCount": {}
        },
        {
            "summary": {},
            "timeCount": {}
        }
    ],
    "test2": [],
    "test3": []
}

```

vishnusomank: Addressed @wazir-ahmed please review

src/observability/input-summary.json
```diff
@@ -0,0 +1,430 @@
+{
+    "Positive test 1": [
```
wazir-ahmed: I don't think the notion of postive/negative cases applies here. The function which is under test, doesn't return a positive/negative value. Its just return some value. Comparing the output of the function with a wrong value doesn't make sense.

src/observability/agg-summary.json
```diff
@@ -0,0 +1,223 @@
+{
+    "Positive test 1": [
```
wazir-ahmed: It is best to put these test input/output files under `test` directory in the root, instead of mixing this with source code.

src/observability/input-summary.json
vishnusomank: The reason for adding the same name on both the input and output array was to ensure that the comparison happens on the correct input/output array. Also positive and negative were just added as keywords to search through and redirect to respective test case functions. Should we change that and keep only one test case?

wazir-ahmed (approved):

src/observability/publisher_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	wd, err := os.Getwd()
+	if err == nil {
+		tempWD := strings.Split(wd, "/")
+		projectHome := strings.Join(tempWD[:len(tempWD)-2], "/")
+
+		inputSummaryFileName = projectHome + "/tests/observability/" + inputSummaryFileName
+		aggSummaryFileName = projectHome + "/tests/observability/" + aggSummaryFileName
+
+	}
```
wazir-ahmed: We can achieve the same result by simply using the relative paths in `inputSummaryFileName` & `aggSummaryFileName`. 

Both are effectively relative paths and going to break if we change the directory hierarchy.

tests/observability/agg-summary.json
```diff
@@ -0,0 +1,118 @@
+{
+    "testcase 1": [
```
wazir-ahmed:
```suggestion
    "test_1": [
```

wazir-ahmed (approved):

seswarrajan (approved):

PR #765, opened 2023-07-04
fix(observability): fix write to db isssue for observability

rajaSahil:
**Purpose of PR?**:

- Fix concurrent write issue which was happening because of using lock in a different goroutine. 
- Fixes write to db for karmor summary

*NOTE*
- There is another issue with protobuf which is causing DE to panic because of invalid go types.

**Checklist:**
- [x] Bug fix. Fixes #<issue number>
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Commit has unit tests

### DIFF
--- a/src/observability/publisher.go
+++ b/src/observability/publisher.go
@@ -20,18 +20,18 @@ func ProcessSystemSummary() {
 	tempSummarizerMap := common.MoveMap(SummarizerMap)
 	SummarizerMapMutex.Unlock()
 
-	var ProcessSystemSummaryWg sync.WaitGroup
+	aggregateSummaryMap(tempSummarizerMap)
 
 	if cfg.GetCfgObservabilityWriteToDB() {
+		var ProcessSystemSummaryWg sync.WaitGroup
 		ProcessSystemSummaryWg.Add(1)
 		go UpsertSummaryCronJob(tempSummarizerMap, &ProcessSystemSummaryWg)
+		ProcessSystemSummaryWg.Wait()
 	}
 
 	if cfg.GetCfgPublisherEnable() {
 		PublisherMutex.Lock()
 		initTime := time.Now()
-
-		aggregateSummaryMap(tempSummarizerMap)
 
 		log.Info().Msgf("Events to publish after aggregation: [%v]", len(tempSummarizerMap))
 		count := 0
@@ -54,7 +54,6 @@ func ProcessSystemSummary() {
 		log.Info().Msgf("Published %v events in %v", count, time.Since(initTime))
 		PublisherMutex.Unlock()
 	}
-	ProcessSystemSummaryWg.Wait()
 }
 
 func aggregateSummaryMap(summaryMap map[types.SystemSummary]types.SysSummaryTimeCount) {


src/observability/publisher.go
```diff
@@ -20,19 +20,19 @@ func ProcessSystemSummary() {
 	tempSummarizerMap := common.MoveMap(SummarizerMap)
 	SummarizerMapMutex.Unlock()
 
-	var ProcessSystemSummaryWg sync.WaitGroup
+	aggregateSummaryMap(tempSummarizerMap)
 
 	if cfg.GetCfgObservabilityWriteToDB() {
+		var ProcessSystemSummaryWg sync.WaitGroup
```
seswarrajan, line 26:
```suggestion
		ProcessSystemSummaryWg := sync.WaitGroup{}
```

seswarrajan (changes requested):

achrefbensaad (approved):
LGTM

src/observability/publisher.go
```diff
@@ -20,19 +20,19 @@ func ProcessSystemSummary() {
...
 		ProcessSystemSummaryWg.Add(1)
 		go UpsertSummaryCronJob(tempSummarizerMap, &ProcessSystemSummaryWg)
+		ProcessSystemSummaryWg.Wait()
```
stefin9898, line 29: We can remove the entire ProcessSystemSummaryWg by removing the call to `UpsertSummaryCronJob` as a goroutine

rajaSahil: I think it will have some performance implications, let's use go routines for this.

seswarrajan:
One more suggestion. Instead of using wg wait and sync, why not write to DB one by one, in line 50, after Publish. We do have a common handle for DB and we are not going to open a new handle every time while we write to DB. 

Will that work ?

Since the logs/summary data is aggreagted, we will have less data and writing to DB in loop individually will not cost much. Please correct me if i am wrong.

src/observability/publisher.go
stefin9898: Since it is waiting at `ProcessSystemSummaryWg.Wait()` then why should there be a goroutine?

stefin9898:
Yes @seswarrajan that make sense! We can just iterate the tempSummarizerMap once and publish / write to db if the flags are enabled for the same

src/observability/publisher.go
rajaSahil: What I  was thinking of is to spawn two go routines for both the task, but I haven't tried to implement it. I will be trying it, so let's have it for now .

rajaSahil:
> One more suggestion. Instead of using wg wait and sync, why not write to DB one by one, in line 50, after Publish. We do have a common handle for DB and we are not going to open a new handle every time while we write to DB.
> 
> Will that work ?
> 
> Since the logs/summary data is aggreagted, we will have less data and writing to DB in loop individually will not cost much. Please correct me if i am wrong.

No, having seperate configs make sense to me!

stefin9898:
> No, having seperate configs make sense to me!

Yes, it will be having separate configs. I believe @seswarrajan meant to iterate the  tempSummarizerMap once. right?

seswarrajan:
> > One more suggestion. Instead of using wg wait and sync, why not write to DB one by one, in line 50, after Publish. We do have a common handle for DB and we are not going to open a new handle every time while we write to DB.
> > Will that work ?
> > Since the logs/summary data is aggreagted, we will have less data and writing to DB in loop individually will not cost much. Please correct me if i am wrong.
> 
> No, having seperate configs make sense to me!

@rajaSahil The config still remains untouched. We will have separate configs. But, we need to call write to db after Publish. For each log. That write will fall inside the condition check.

rajaSahil:
@seswarrajan Unable to get you. Why not we spawn two different goroutines and have it processed at the same time?

src/observability/publisher.go
```diff
@@ -20,19 +20,19 @@ func ProcessSystemSummary() {
 	tempSummarizerMap := common.MoveMap(SummarizerMap)
 	SummarizerMapMutex.Unlock()
 
-	var ProcessSystemSummaryWg sync.WaitGroup
+	aggregateSummaryMap(tempSummarizerMap)
```
vishnusomank, line 23: Do we really need to aggregate and store in the DB? IMO we should keep the raw data in DB and use agg before publishing it to SaaS? WDYT @rajaSahil @seswarrajan

rajaSahil: We are anyway aggregating  the data before pushing to SAAS, why can't we store aggregated result in the DB? Is it going to create any issues? We anyway aggregate all the file data for karmor summary, so we can do 1st level of aggregation over here which can increase performance for karmor summary also.

vishnusomank: It is not going to create any issues. And the agg is false by default in karmor summary right? That is why I suggested to keep it raw.

Ankurk99 (approved):
LGTM!

seswarrajan (approved):

PR #775, opened 2023-09-11
fix(hardening): Download policy-template zip

PaveenV:
**Issue:** grab.Get throws error 403 Forbidden while downloading the zip file

**Fix:** Removed the grab package dependency and used net/http for downloading the policy-template zip file.

**Ticket:** https://accu-knox.atlassian.net/browse/CNAPP-7814

### DIFF
--- a/src/recommendpolicy/downloadTemplates.go
+++ b/src/recommendpolicy/downloadTemplates.go
@@ -5,6 +5,7 @@ import (
 	"archive/zip"
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -12,7 +13,6 @@ import (
 	"strings"
 
 	"github.com/accuknox/auto-policy-discovery/src/types"
-	"github.com/cavaliergopher/grab/v3"
 	"github.com/google/go-github/github"
 	"sigs.k8s.io/yaml"
 )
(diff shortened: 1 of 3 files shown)


src/recommendpolicy/downloadTemplates.go
```diff
@@ -87,6 +88,37 @@ func init() {
...
+
+	out, err := os.Create(filepath.Clean(destination))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := out.Close(); err != nil {
+			log.Error().Msgf("Error while closing os file %s\n", err)
+		}
+	}()
```
vishnusomank:
```suggestion
    resp, err := http.Get(url)
    if err != nil {
        return err
    }
    defer resp.Body.Close()

    out, err := os.Create(filepath.Clean(destination))
    if err != nil {
        return err
    }
    defer out.Close()
```

vishnusomank (changes requested):

src/recommendpolicy/downloadTemplates.go
PaveenV: addressed please verify

vishnusomank (approved):
LGTM!

wazir-ahmed (approved):

